PR #15, opened 2019-11-28
BLS crypto provider

sergkaprovich:
### BLS signature scheme
- Generate key pair;
- Derive public key from private;
- Generate signature;
- Verify signature;

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,10 @@ jobs:
           - os: macOS-10.14
             compiler:
               cc: gcc
+        include:
+          - os: macOS-10.14
+            env:
+              CFLAGS: "-I /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include"
     steps:
     - uses: actions/checkout@v1
       name: checkout
@@ -34,6 +38,7 @@ jobs:
         submodules: true
         clean: true
         fetch-depth: 1
+        token: ${{ secrets.CI_TOKEN }}
     - name: install dependencies
       run: |
         set -e
@@ -46,6 +51,7 @@ jobs:
 
         sudo python3 -m pip install --upgrade pip
         sudo pip3 install scikit-build cmake requests gitpython gcovr pyyaml
+        sudo curl https://sh.rustup.rs -sSf | sh -s -- -y
     - name: cmake
       env:
         CC: ${{ matrix.compiler.cc }}
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "deps/bls-signatures"]
+	path = deps/bls-signatures
+	url = https://github.com/filecoin-project/bls-signatures.git
--- a/core/crypto/bls_provider/bls_provider.hpp
+++ b/core/crypto/bls_provider/bls_provider.hpp
@@ -1,0 +1,57 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef CRYPTO_BLS_PROVIDER_HPP
+#define CRYPTO_BLS_PROVIDER_HPP
+
+#include <gsl/span>
+
+#include "crypto/bls_provider/bls_types.hpp"
+
+namespace fc::crypto::bls {
+  /**
+   * @class BLS provider - C++ wrapper around Rust BLS implementation
+   */
+  class BlsProvider {
+   public:
+    virtual ~BlsProvider() = default;
+    /**
+     * @brief Generate BLS key pair
+     * @return BLS private and public keys or error code
+     */
+    virtual outcome::result<KeyPair> generateKeyPair() const = 0;
+
+    /**
+     * @brief Generate BLS public key from private
+     * @param key - private BLS key
+     * @return public BLS key or error code
+     */
+    virtual outcome::result<PublicKey> derivePublicKey(
+        const PrivateKey &key) const = 0;
+
+    /**
+     * @brief Generate BLS signature
+     * @param message - data to sign
+     * @param key - BLS private key
+     * @return BLS signature or error code
+     */
+    virtual outcome::result<Signature> sign(gsl::span<uint8_t> message,
+                                            const PrivateKey &key) const = 0;
+
+    /**
+     * @brief Verify BLS signature
+     * @param message - signed data
+     * @param signature - BLS signature for verifying
+     * @param key - BLS public key
+     * @return signature status or error code
+     */
+    virtual outcome::result<bool> verifySignature(
+        gsl::span<uint8_t> message,
+        const Signature &signature,
+        const PublicKey &key) const = 0;
+  };
+}  // namespace fc::crypto::bls
+
+#endif
--- a/core/crypto/bls_provider/bls_types.hpp
+++ b/core/crypto/bls_provider/bls_types.hpp
@@ -1,0 +1,41 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef CRYPTO_BLS_PROVIDER_TYPES_HPP
+#define CRYPTO_BLS_PROVIDER_TYPES_HPP
+
+#include <array>
+
+extern "C" {
+#include "crypto/bls_provider/libbls_signatures.h"
+}
+
+#include "common/outcome.hpp"
+
+namespace fc::crypto::bls {
+
+  using PrivateKey = std::array<uint8_t, PRIVATE_KEY_BYTES>;
+  using PublicKey = std::array<uint8_t, PUBLIC_KEY_BYTES>;
+  using Signature = std::array<uint8_t, SIGNATURE_BYTES>;
+  using Digest = std::array<uint8_t, DIGEST_BYTES>;
+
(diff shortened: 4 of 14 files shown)


.github/workflows/ci.yml
```diff
@@ -46,13 +47,15 @@ jobs:
...
         sudo pip3 install scikit-build cmake requests gitpython gcovr pyyaml
+        sudo curl https://sh.rustup.rs -sSf | sh -s -- -y
     - name: cmake
       env:
         CC: ${{ matrix.compiler.cc }}
         CXX: ${{ matrix.compiler.cxx }}
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # has to be included to access other secrets
         GITHUB_HUNTER_USERNAME: ${{ secrets.GITHUB_HUNTER_USERNAME }}
         GITHUB_HUNTER_TOKEN: ${{ secrets.GITHUB_HUNTER_TOKEN }}
+        CFLAGS: "-I /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include"
```
sergkaprovich: Possible point to refactor:
needed for compiling of the Rust system libraries, without this environment variable we will not be able to find C headers (such as errno.h and so on) on MacOS. Linux builds are successful.

core/crypto/bls_provider/libbls_signatures.h
```diff
@@ -0,0 +1,126 @@
+/* libbls_signatures Header Version 0.1.0 */
+
```
sergkaprovich: This file is generated while compiling Rust implementation of the BLS signatures and we don't need to copy it here except one reason: we can't suppress clang-tidy warning about "Include header not found".
Also add Rust toolchain to the clang-tidy CI/CD and build project again isn't a good decision.

core/crypto/bls_provider/bls_provider.hpp
```diff
@@ -0,0 +1,58 @@
...
+     * @param signature - BLS signature for verifying
+     * @param key - BLS public key
+     * @return signature status or error code
+     */
+    virtual outcome::result<bool> VerifySignature(
+        gsl::span<uint8_t> message,
+        const Signature &signature,
+        const PublicKey &key) const = 0;
+
+    virtual ~BlsProvider() = default;
```
Alexey-N-Chernyshov: https://google.github.io/styleguide/cppguide.html#Declaration_Order

core/crypto/bls_provider/bls_provider_impl.hpp
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef CRYPTO_BLS_PROVIDER_IMPL_HPP
+#define CRYPTO_BLS_PROVIDER_IMPL_HPP
+
+#include "crypto/bls_provider/bls_provider.hpp"
+
+namespace fc::crypto::bls {
+  class BlsProviderImpl : public BlsProvider {
```
Alexey-N-Chernyshov: Please add comments.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+
```
kamilsa:
```suggestion
```

.gitmodules
```diff
@@ -0,0 +1,4 @@
...
+[submodule "deps/bls-signatures"]
+	path = deps/bls-signatures
+	url = https://github.com/filecoin-project/bls-signatures.git
```
kamilsa: can we fix commit somehow, so that when bls-signatures updated our implementation does not crash

CMakeLists.txt
```diff
@@ -122,3 +122,15 @@ if (TESTING)
   enable_testing()
   add_subdirectory(test)
 endif ()
+
+# update git submodules
+find_package(Git QUIET)
```
kamilsa: Why do we need to update submodules on every cmake run? 
I think it is enough to clone some certain commit of remote repo and update submodule as separate PR to cpp-filecoin when required.
Or I didn't get what this cmake is doing

core/crypto/bls_provider/bls_provider_impl.cpp
```diff
@@ -0,0 +1,113 @@
...
+      gsl::span<uint8_t> message,
+      const Signature &signature,
+      const PublicKey &key) const {
+    OUTCOME_TRY(digest, GenerateHash(message));
+    if (verify(signature.data(),
+               digest.data(),
+               digest.size(),
+               key.data(),
+               key.size())
+        == 1) {
```
kamilsa: Minor issue: usually rule is: 0 is false, everything else is true. I think here it is also the case
```suggestion
        != 0) {
```

core/crypto/bls_provider/bls_provider_impl.cpp
```diff
@@ -0,0 +1,113 @@
...
+      return outcome::success(true);
```
kamilsa: this does not work?
```suggestion
      return true;
```

core/crypto/bls_provider/bls_provider_impl.hpp
kamilsa: TBH I don't think we need comments like:
`// Class implementing BlsProvider`
It is enough to have comments only for BlsProvider unless BlsProviderImpl contains some new methods that are not inherited from the parent class

core/crypto/bls_provider/bls_provider_impl.hpp
```diff
@@ -0,0 +1,37 @@
...
+   public:
+    outcome::result<KeyPair> GenerateKeyPair() const override;
```
kamilsa: Here and in other place please always use camelCase from small letter for methods names

core/crypto/bls_provider/bls_types.hpp
```diff
@@ -0,0 +1,37 @@
...
+#ifndef CRYPTO_BLS_PROVIDER_TYPES_HPP
+#define CRYPTO_BLS_PROVIDER_TYPES_HPP
+
+#include <array>
+
+#include "common/outcome.hpp"
+
+namespace fc::crypto::bls {
+
+  using PrivateKey = std::array<uint8_t, 32>;
```
kamilsa: are there constants with private key, public key and signature sizes from bls library? If so can we use them instead of magic numbers?

kamilsa: There are some:
https://github.com/filecoin-project/cpp-filecoin/pull/15/files#diff-16aff5edd15d86e03435647aa6164dabR8-R14. Use them instead of magic numbers

test/core/crypto/bls_provider_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "crypto/bls_provider/bls_provider_impl.hpp"
+
+#include <vector>
+#include <algorithm>
+
+#include <gtest/gtest.h>
+#include <testutil/outcome.hpp>
```
kamilsa:
```suggestion
#include "testutil/outcome.hpp"
```

CMakeLists.txt
sergkaprovich: It's just for convenience - we should call "git submodule update" to download submodules content after Filecoin repo clone, maybe, it's unnecessarily?

.gitmodules
sergkaprovich: It's fixed in the "deps/bls-signatures" file: `Submodule bls-signatures added at 92e4c4` (commit 92e4c48c12bf575191903018f3bfc4a53fc73023)

core/crypto/bls_provider/bls_provider_impl.cpp
```diff
@@ -0,0 +1,109 @@
...
+  outcome::result<KeyPair> BlsProviderImpl::generateKeyPair() const {
+    KeyPair key_pair{};
+    PrivateKeyGenerateResponse *private_key_response = private_key_generate();
+    if (private_key_response == nullptr) {
+      return Errors::KeyPairGenerationFailed;
+    }
+    auto free_private_response = gsl::finally([private_key_response]() {
+      destroy_private_key_generate_response(private_key_response);
+    });
+    std::move(std::begin(private_key_response->private_key),
```
masterjedy: moving bytes is questionable

sergkaprovich: Change to std::copy?

CMakeLists.txt
masterjedy: I don't think you need it.

.github/workflows/ci.yml
masterjedy: Is there a way how to add that directory only when matrix.os is macOS?

CMakeLists.txt
masterjedy: If it works without this step, pls remove it. We should be forced to update submodules like that each time cmake reruns. What if internet connection is unstable?

.github/workflows/ci.yml
sergkaprovich: Found solution to add that directory only for macOS

core/crypto/bls_provider/bls_provider_impl.cpp
```diff
@@ -0,0 +1,109 @@
+/**
```
kamilsa: Please put impl files into /impl folder (`core/crypto/bls/impl`)
Sorry for not finding this during first review

kamilsa: also folder should be consistent with namespace name

PR #16, opened 2019-11-29
FIL-21 Address component implementation

ekovalev:

### Description of the Change
Filecoin Address component implementation

### Benefits
Numerous

### Possible Drawbacks 
None

### DIFF
--- a/core/primitives/address.cpp
+++ b/core/primitives/address.cpp
@@ -13,9 +13,11 @@ OUTCOME_CPP_DEFINE_CATEGORY(fc::primitives, AddressError, e) {
     case (AddressError::UNKNOWN_PROTOCOL):
       return "Failed to create address: unknown address protocol";
     case (AddressError::INVALID_PAYLOAD):
-      return "Failed to create address: invalid payload for the specified protocol";
+      return "Failed to create address: invalid payload for the specified "
+             "protocol";
     case (AddressError::UNKNOWN_NETWORK):
-      return "Failed to create address: network must either be mainnet or testnet";
+      return "Failed to create address: network must either be MAINNET or "
+             "TESTNET";
     default:
       return "Failed to create address: unknown error";
   };
@@ -23,33 +25,34 @@ OUTCOME_CPP_DEFINE_CATEGORY(fc::primitives, AddressError, e) {
 
 namespace fc::primitives {
 
-  using common::Blob;
-
-  bool Address::IsKeyType() const {
-    return visit_in_place(data_,
+  bool Address::isKeyType() const {
+    return visit_in_place(data,
                           [](uint64_t v) { return false; },
                           [](const Secp256k1PublicKeyHash &v) { return true; },
                           [](const ActorExecHash &v) { return false; },
                           [](const BLSPublicKeyHash &v) { return true; });
   }
 
-  Protocol Address::GetProtocol() const {
-    return visit_in_place(data_,
-                          [](uint64_t v) { return Protocol::ID; },
-                          [](const Secp256k1PublicKeyHash &v) { return Protocol::SECP256K1; },
-                          [](const ActorExecHash &v) { return Protocol::Actor; },
-                          [](const BLSPublicKeyHash &v) { return Protocol::BLS; });
+  Protocol Address::getProtocol() const {
+    return visit_in_place(
+        data,
+        [](uint64_t v) { return Protocol::ID; },
+        [](const Secp256k1PublicKeyHash &v) { return Protocol::SECP256K1; },
+        [](const ActorExecHash &v) { return Protocol::ACTOR; },
+        [](const BLSPublicKeyHash &v) { return Protocol::BLS; });
   }
 
   bool operator==(const Address &lhs, const Address &rhs) {
-    return lhs.network_ == rhs.network_ && lhs.GetProtocol() == rhs.GetProtocol() && lhs.data_ == rhs.data_;
+    return lhs.network == rhs.network && lhs.getProtocol() == rhs.getProtocol()
+           && lhs.data == rhs.data;
   }
 
   bool operator<(const Address &lhs, const Address &rhs) {
-    return lhs.network_ < rhs.network_
-           || (lhs.network_ == rhs.network_
-               && (lhs.GetProtocol() < rhs.GetProtocol()
-                   || (lhs.GetProtocol() == rhs.GetProtocol() && lhs.data_ < rhs.data_)));
+    return lhs.network < rhs.network
+           || (lhs.network == rhs.network
+               && (lhs.getProtocol() < rhs.getProtocol()
+                   || (lhs.getProtocol() == rhs.getProtocol()
+                       && lhs.data < rhs.data)));
   }
 
 };  // namespace fc::primitives
--- a/core/primitives/address.cpp
+++ b/core/primitives/address.cpp
@@ -5,6 +5,51 @@
 
 #include "primitives/address.hpp"
 
-namespace fc::primitives {
-  // implementation
+#include "common/visitor.hpp"
+
+OUTCOME_CPP_DEFINE_CATEGORY(fc::primitives, AddressError, e) {
+  using fc::primitives::AddressError;
+  switch (e) {
+    case (AddressError::UNKNOWN_PROTOCOL):
+      return "Failed to create address: unknown address protocol";
+    case (AddressError::INVALID_PAYLOAD):
+      return "Failed to create address: invalid payload for the specified protocol";
+    case (AddressError::UNKNOWN_NETWORK):
+      return "Failed to create address: network must either be mainnet or testnet";
+    default:
+      return "Failed to create address: unknown error";
+  };
 }
+
+namespace fc::primitives {
+
+  using common::Blob;
+
+  bool Address::IsKeyType() const {
+    return visit_in_place(data_,
+                          [](uint64_t v) { return false; },
+                          [](const Secp256k1PublicKeyHash &v) { return true; },
+                          [](const ActorExecHash &v) { return false; },
+                          [](const BLSPublicKeyHash &v) { return true; });
+  }
+
+  Protocol Address::GetProtocol() const {
+    return visit_in_place(data_,
+                          [](uint64_t v) { return Protocol::ID; },
+                          [](const Secp256k1PublicKeyHash &v) { return Protocol::SECP256K1; },
+                          [](const ActorExecHash &v) { return Protocol::Actor; },
+                          [](const BLSPublicKeyHash &v) { return Protocol::BLS; });
+  }
+
+  bool operator==(const Address &lhs, const Address &rhs) {
+    return lhs.network_ == rhs.network_ && lhs.GetProtocol() == rhs.GetProtocol() && lhs.data_ == rhs.data_;
+  }
+
+  bool operator<(const Address &lhs, const Address &rhs) {
+    return lhs.network_ < rhs.network_
+           || (lhs.network_ == rhs.network_
+               && (lhs.GetProtocol() < rhs.GetProtocol()
+                   || (lhs.GetProtocol() == rhs.GetProtocol() && lhs.data_ < rhs.data_)));
+  }
+
+};  // namespace fc::primitives
--- a/core/primitives/address.hpp
+++ b/core/primitives/address.hpp
@@ -3,16 +3,85 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#ifndef CMAKE_HUNTER_SEED_CORE_ADDRESS_ADDRESS_HPP
-#define CMAKE_HUNTER_SEED_CORE_ADDRESS_ADDRESS_HPP
+#ifndef CPP_FILECOIN_CORE_PRIMITIVES_ADDRESS_HPP
+#define CPP_FILECOIN_CORE_PRIMITIVES_ADDRESS_HPP
+
+#include <cstdint>
+
+#include <boost/variant.hpp>
+#include "common/blob.hpp"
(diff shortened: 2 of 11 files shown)


core/primitives/address.hpp
```diff
@@ -3,16 +3,146 @@
...
+    // implemented, as well?
+    String encode() {
+      return this->string();
+    }
+    bool validateChecksum(Bytes checksum) {
+      return true;
+    }
+
+   private:
+    Payload data;
```
Alexey-N-Chernyshov: please, adhere 
https://google.github.io/styleguide/cppguide.html#Variable_Names
e.g. a_class_data_member_

core/primitives/address.hpp
```diff
@@ -3,16 +3,146 @@
...
+
+    // Older API defined in
+    // https://filecoin-project.github.io/specs/#appendix__address Should it be
+    // implemented, as well?
+    String encode() {
+      return this->string();
+    }
+    bool validateChecksum(Bytes checksum) {
+      return true;
+    }
```
Alexey-N-Chernyshov: Can you add operator <, please? So, it van be used in maps.
Operator == also will be helpful.

ekovalev: Bytes is an alias for std::vector<uint8_t>, therefore the == operator comes for free. Noted on < though

core/primitives/address.hpp
```diff
@@ -3,16 +3,146 @@
...
+
+    Bytes toBytes();
+
    public:
-    //TODO (kamilsa): FIL-21 Implement primitives
+    // Factory method to create Address object
+    static outcome::result<std::unique_ptr<Address>> makeAddress(
+        Network net, Byte protocol, const Bytes &payload);
+
+    // TODO: Implement creation of an Address by decoding a String (if
```
kamilsa: For every TODO please create the task in Jira, so that we do not forget about that

core/primitives/address.hpp
```diff
@@ -3,16 +3,146 @@
...
+
+    const Network getNetwork() {
+      return networkId;
+    }
+    const bool isEmpty() {
+      return empty;
+    }
+    const Payload &getData() {
+      return data;
+    };
```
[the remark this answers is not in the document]
kamilsa: same for other methods in hpp

test/core/primitives/CMakeLists.txt
```diff
@@ -8,4 +8,5 @@ addtest(address_test
     )
 target_link_libraries(address_test
     address
+    leb128
```
kamilsa: If you use leb128 in address then you should link it to address instead of address test

core/primitives/address.cpp
```diff
@@ -4,7 +4,170 @@
  */
 
 #include "primitives/address.hpp"
```
kamilsa: add empty line after this include. (Check google code style for the order of includes)

core/primitives/address.cpp
```diff
@@ -4,7 +4,170 @@
...
+#include "codec/leb128/leb128.hpp"
+#include "common/visitor.hpp"
+#include "cppcodec/base32_default_rfc4648.hpp"
```
kamilsa: this external library. Should be in <> instead ""

core/primitives/address.cpp
```diff
@@ -4,7 +4,170 @@
...
+#include "crypto/blake2/blake2b.h"
 
-namespace fc::primitives {
-  // implementation
+#include <memory>
```
kamilsa: wrong place. 
Proper order of includes:
```
#include "primitives/address.hpp"

#include <memory>

#include <cppcodec/base32_default_rfc4648.hpp>
#include "codec/leb128/leb128.hpp"
#include "common/visitor.hpp"
#include "crypto/blake2/blake2b.h"
```

core/primitives/address.hpp
Alexey-N-Chernyshov: I mean for Address class.

core/primitives/address.hpp
```diff
@@ -3,16 +3,146 @@
...
+#include <type_traits>
+#include <vector>
+
+#include "boost/variant.hpp"
+#include "common/outcome.hpp"
 
 namespace fc::primitives {
+  using Uint = uint64_t;            // TODO:
+  using Byte = uint8_t;             // consider moving
+  using Bytes = std::vector<Byte>;  // commonly used types
```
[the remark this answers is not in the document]
ekovalev: Fair enough. It's just what they are called in the spec so arguably it could be easier to match things up. But it didn't seem right still, hence the "TODO" comment

core/primitives/address.hpp
ekovalev: This is a consequence of ambiguous spec for the Address component which is introduced in two places in the document itself and has the 3rd "source of truth" in the reference implementation. Until the spec has been completely solidified we won't know what the final interface is going to be. It is possible that this method will just go because it is not in the most latest spec variation and I don't see a use case for it. I'm not sure a Jira task is appropriate

test/core/primitives/CMakeLists.txt
ekovalev: I think the test wouldn't link without that but I can be wrong - will check

core/primitives/address.cpp
ekovalev: Sure. Will check what google has to say about this aspect of code style

core/primitives/address.hpp
```diff
@@ -3,16 +3,146 @@
...
+  enum class Network : int { testnet = 0, mainnet };
+
+  enum class Protocol : int { ID = 0, SECP256K1, Actor, BLS };
+
+  template <typename T>
+  constexpr auto to_underlying(T v) noexcept {
+    return static_cast<std::underlying_type_t<T>>(v);
+  }
+
+  struct PublicKeySecp256k1Hash {
```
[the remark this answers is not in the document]
ekovalev: This is merely a wrapper around std::vector<uint8_t> - a necessary workaround for the boost::variant to be able to differentiate between the alternatives of the same type (byte array). It's not intended to be public; the public interface will always take byte array as an input.

Furthermore, a need for such complication gives rise to a question whether a tagged union is the best way to implement 4 types of Addresses or a simple inheritance from an abstract base Address class would be more suitable. I think there is still room for discussion here because either of the approaches has their pros and cons

core/primitives/address.hpp
kamilsa: I am afraid this will be lost without task

test/core/primitives/CMakeLists.txt
kamilsa: yes. To make it link you need to link address library with leb128. 
Otherwise, every dependency of address would need to be linked with leb128 manually, whereas it could be avoided by just linking address with lev128

core/primitives/address.cpp
```diff
@@ -5,6 +5,163 @@
...
+        BLSPublicKeyHash hash{};
+        std::copy_n(payload.begin(), kBLSHashSize, hash.begin());
+        return outcome::success(std::unique_ptr<Address>(new Address(net, hash)));
+      }
+      default:
+        return outcome::failure(AddressError::UNKNOWN_PROTOCOL);
+    }
+  }
+
+  bool operator==(const Address &lhs, const Address &rhs) {
```
Alexey-N-Chernyshov: Should we also check protocol? I understand that the same hash is almost impossible, but
f2abc == f3abc

core/primitives/address.hpp
Alexey-N-Chernyshov: Thanks a lot!

core/primitives/address.cpp
```diff
@@ -5,6 +5,163 @@
...
+  Address::Address(Network net, uint64_t id) : data_(id), network_(net) {}
+  Address::Address(Network net, Secp256k1PublicKeyHash &&payload) noexcept : data_(payload), network_(net) {}
+  Address::Address(Network net, const Secp256k1PublicKeyHash &payload) : data_(payload), network_(net) {}
+  Address::Address(Network net, ActorExecHash &&payload) noexcept : data_(payload), network_(net) {}
+  Address::Address(Network net, const ActorExecHash &payload) : data_(payload), network_(net) {}
+  Address::Address(Network net, BLSPublicKeyHash &&payload) noexcept : data_(payload), network_(net) {}
+  Address::Address(Network net, const BLSPublicKeyHash &payload) : data_(payload), network_(net) {}
+
+  bool Address::VerifySyntax() {
+    // Always return true for now as all addresses are syntactically correct by default
```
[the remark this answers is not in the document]
ekovalev: The checks are done in the makeAddress function - that's the only way to create an Address that makes sense. Once created, an address is guaranteed to be syntactically correct. Am I wrong?

core/primitives/address.cpp
ekovalev: An idea like that has come across my mind, of course but what are the odds? Anyway, checking the protocol is a small overhead after all, so I'll just add it

core/primitives/address.cpp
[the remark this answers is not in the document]
Alexey-N-Chernyshov: I see. In this case I see no reason why we should keep this function.

core/primitives/address.hpp
```diff
@@ -3,16 +3,85 @@
...
+  enum class AddressError {
+    UNKNOWN_PROTOCOL = 1, /**< Unknown Address protocol/type */
+    INVALID_PAYLOAD,      /**< Invalid data for a given protocol */
+    UNKNOWN_NETWORK       /**< Unknown network: neither testnet nor mainnet */
+  };
+
+  /**
+   * @brief Supported networks inside which addresses make sense
+   */
+  enum Network : uint8_t { mainnet = 0x0, testnet = 0x1 };
```
kamilsa: please update naming according google style: v

core/primitives/address.cpp
```diff
@@ -5,6 +5,51 @@
...
+    case (AddressError::UNKNOWN_NETWORK):
+      return "Failed to create address: network must either be mainnet or testnet";
+    default:
+      return "Failed to create address: unknown error";
+  };
 }
+
+namespace fc::primitives {
+
+  using common::Blob;
```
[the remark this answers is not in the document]
ekovalev: Not anymore - a leftover from a previous version. Removed

core/primitives/address_codec.cpp
```diff
@@ -0,0 +1,214 @@
...
+                              return ingest;
+                            },
+                            [&ingest](const ActorExecHash &v) {
+                              ingest.insert(ingest.end(), v.begin(), v.end());
+                              return ingest;
+                            },
+                            [&ingest](const BLSPublicKeyHash &v) {
+                              ingest.insert(ingest.end(), v.begin(), v.end());
+                              return ingest;
+                            });
```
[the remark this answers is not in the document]
ekovalev: Nice hack - didn't know that. Will apply wherever possible

core/primitives/address_codec.cpp
```diff
@@ -0,0 +1,214 @@
...
+        [](const BLSPublicKeyHash &v) { return std::vector<uint8_t>(v.begin(), v.end()); });
+
+    std::vector chksum = checksum(address);
+
+    payload.insert(payload.end(), chksum.begin(), chksum.end());
+    auto encoded = base32::encode(payload);
+
+    // Convert to lower case, as the spec requires: https://filecoin-project.github.io/specs/#payload
+    std::transform(
+        encoded.begin(), encoded.end(), encoded.begin(), [](unsigned char c) { return std::tolower(c); });
```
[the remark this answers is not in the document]
ekovalev: yes. It's just that I set the line length to 110. As far as I know there is no common opinion on that yet. However, I agree it looks ugly here - probably better revert and do 80

core/primitives/address_codec.cpp
```diff
@@ -0,0 +1,216 @@
...
+          return outcome::failure(AddressError::INVALID_PAYLOAD);
+        }
+        return outcome::success(Address{net, decoded.value()});
+      }
+      case Protocol::SECP256K1: {
+        if (payload.size() != kBlake2b160HashSize) {
+          return outcome::failure(AddressError::INVALID_PAYLOAD);
+        }
+        Secp256k1PublicKeyHash hash{};
+        std::copy_n(payload.begin(), kBlake2b160HashSize, hash.begin());
```
Alexey-N-Chernyshov:
```suggestion
        std::copy_n(std::make_move_iterator(payload.begin()), kBlake2b160HashSize, hash.begin());
```
Move instead copy.
Here and further.

core/primitives/address_codec.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+    auto p = static_cast<Protocol>(*(pos++));
+    std::vector<uint8_t> payload(pos, v.end());
+    switch (p) {
+      case Protocol::ID: {
+        auto decoded = codec::leb128::decode<uint64_t>(payload);
+        if (!decoded) {
+          return outcome::failure(AddressError::INVALID_PAYLOAD);
+        }
+        return outcome::success(Address{net, decoded.value()});
```
Alexey-N-Chernyshov:
```
case Protocol::ID: {
        OUTCOME_TRY(decoded, codec::leb128::decode<uint64_t>(payload));
        return Address{net, decoded};
}
```
Here and further

Alexey-N-Chernyshov: * After pondering over. *
There is almost no sense to move basic types, but still move is always better.

core/primitives/address_codec.cpp
Alexey-N-Chernyshov: * After pondering over. *
There is almost no sense to move basic types, but still move is always better.

PR #17, opened 2019-12-05
Fix LEB128 decode overflow

sergkaprovich:
In previous version decode overflow was detected too late:
uint64_t max value is 2^64-1, it's LEB128 representation is 
`{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x01}`

Next value is 2^64, which must cause overflow error:
`{0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x02}`

But overflow for uint64_t was detected only on values, greater than 2^70.

I don't think, that we need separate test for each type (uint8_t, uint16_t ...), because overflow check is common for all data types (uses uint8_t for check execution)

### DIFF
--- a/core/codec/leb128/README.md
+++ b/core/codec/leb128/README.md
@@ -34,6 +34,6 @@ using fc::codec::leb128::encode;
 using fc::codec::leb128::decode;
 
 T data;
-OUTCOME_TRY(encoded, encode(data));
+std::vector<uint8_t> encoded = encode(data);
 OUTCOME_TRY(decoded, decode(encoded));
 ```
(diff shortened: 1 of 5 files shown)


core/codec/leb128/README.md
```diff
@@ -1,32 +1,40 @@
-# LEB128 codec
-> Primitive types encode/decode user guide
+# LEB128 codec C++ implementation
```
kamilsa: Add section about custom types

PR #23, opened 2019-12-10
RLE+ codec implementation

sergkaprovich:
### Description of the Change

RLE+ data codec

[Reference Specification](https://filecoin-project.github.io/specs/#rle-bitset-encoding)

### Benefits

Ability to provide RLE+ encode/decode features

### Possible Drawbacks 

None

### DIFF
--- a/core/codec/rle/CMakeLists.txt
+++ b/core/codec/rle/CMakeLists.txt
@@ -1,0 +1,9 @@
+add_library(rle_plus_codec
+    rle_plus_encoding_stream.cpp
+    rle_plus_errors.cpp
+    )
+
+target_link_libraries(rle_plus_codec
+    Boost::boost
+    outcome
+  )
--- a/core/codec/rle/rle_plus.hpp
+++ b/core/codec/rle/rle_plus.hpp
@@ -1,0 +1,54 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef CODEC_RLE_PLUS_HPP
+#define CODEC_RLE_PLUS_HPP
+
+#include "common/outcome.hpp"
+#include "codec/rle/rle_plus_errors.hpp"
+#include "codec/rle/rle_plus_decoding_stream.hpp"
+#include "codec/rle/rle_plus_encoding_stream.hpp"
+
+namespace fc::codec::rle {
+  /**
+   * @brief RLE+ encode
+   * @tparam T - type of elements to encode
+   * @tparam A - std::set allocator
+   * @param input - data to encode
+   * @return Encoded byte-vector
+   */
+  template <typename T, typename A>
+  std::vector<uint8_t> encode(const std::set<T, A> &input) {
+    RLEPlusEncodingStream encoder;
+    encoder << input;
+    return encoder.data();
+  }
+
+  /**
+   * @brief RLE+ decode
+   * @tparam T - type of elements to decode
+   * @param input - data to decode
+   * @return Decoded data
+   */
+  template <typename T>
+  outcome::result<std::set<T>> decode(gsl::span<const uint8_t> input) {
+    std::set<T> data;
+    RLEPlusDecodingStream decoder(input);
+    try {
+      decoder >> data;
+    } catch (errors::VersionMismatch&) {
+      return RLEPlusDecodeError::VersionMismatch;
+    } catch (errors::IndexOutOfBound&) {
+      return RLEPlusDecodeError::DataIndexFailure;
+    } catch (errors::UnpackBytesOverflow&) {
+      return RLEPlusDecodeError::UnpackOverflow;
+    } catch (errors::MaxSizeExceed&) {
+      return RLEPlusDecodeError::MaxSizeExceed;
+    }
+    return data;
+  }
+};  // namespace fc::codec::rle
+
+#endif
--- a/core/codec/rle/rle_plus_decoding_stream.hpp
+++ b/core/codec/rle/rle_plus_decoding_stream.hpp
@@ -1,0 +1,180 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef CPP_FILECOIN_RLE_PLUS_DECODING_STREAM_HPP
+#define CPP_FILECOIN_RLE_PLUS_DECODING_STREAM_HPP
+
+#include <vector>
+#include <set>
+
+#include <boost/dynamic_bitset.hpp>
+#include <gsl/span>
+
+#include "codec/rle/rle_plus_config.hpp"
+#include "codec/rle/rle_plus_errors.hpp"
+
+namespace fc::codec::rle {
+  /**
+   * @class Decode RLE+ byte stream
+   */
+  class RLEPlusDecodingStream {
+   public:
+    /**
+     * @brief Constructor
+     * @param data - RLE+ encoded bytes
+     */
+    explicit RLEPlusDecodingStream(gsl::span<const uint8_t> data)
+        : index_{}, magnitude_{}, content_{data.begin(), data.end()} {}
+
+    /**
+     * @brief Decode RLE+
+     * @tparam T - type of output data
+     * @param output - decoded data
+     * @return Decoded stream
+     */
+    template <typename T>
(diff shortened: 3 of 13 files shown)


core/codec/rle/rle_plus.hpp
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  /**
+   * @brief RLE+ decode
+   * @tparam T - type of elements to decode
+   * @param input - data to decode
+   * @return Decoded data
+   */
+  template <typename T>
+  outcome::result<std::set<T>> RLEPlusDecode(gsl::span<const uint8_t> input) {
```
Alexey-N-Chernyshov: Should `input` be passed by const ref?

core/codec/rle/rle_plus_decoding_stream.hpp
```diff
@@ -0,0 +1,180 @@
...
+  /**
+   * @class Decode RLE+ byte stream
+   */
+  class RLEPlusDecodingStream {
+   public:
+    /**
+     * @brief Constructor
+     * @param data - RLE+ encoded bytes
+     */
+    explicit RLEPlusDecodingStream(gsl::span<const uint8_t> data)
```
Alexey-N-Chernyshov: const ref for data

core/codec/rle/rle_plus_decoding_stream.hpp
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /**
+     * @brief Decode single RLE+ block
+     * @tparam T - type of the data to output
+     * @param current_value - index of the current number
+     * @param output - decoded data
+     */
+    template <typename T>
+    void decodeSingleBlock(T &current_value, std::set<T> &output) {
```
Alexey-N-Chernyshov: Prefer using return values over output parameters.

core/codec/rle/rle_plus_decoding_stream.hpp
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /**
+     * @brief Decode small RLE+ block
+     * @tparam T - type of the data to output
+     * @param current_value - index of the current number
+     * @param output - decoded data
+     */
+    template <typename T>
+    void decodeSmallBlock(T &current_value, std::set<T> &output) {
```
Alexey-N-Chernyshov: const T &current_value

Alexey-N-Chernyshov: Prefer using return values over output parameters.

core/codec/rle/rle_plus_decoding_stream.hpp
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /**
+     * @brief Decode long RLE+ block
+     * @tparam T - type of the data to output
+     * @param current_value - index of the current number
+     * @param output - decoded data
+     */
+    template <typename T>
+    void decodeLongBlock(T &current_value, std::set<T> &output) {
```
Alexey-N-Chernyshov: sap

core/codec/rle/rle_plus_encoding_stream.hpp
```diff
@@ -0,0 +1,143 @@
...
+        }
+      }
+      return *this;
+    }
+
+    /**
+     * @brief Get encoded stream content
+     * @return Stream content
+     */
+    std::vector<uint8_t> data() const {
```
Alexey-N-Chernyshov: Please, move function body to .cpp.

core/codec/rle/rle_plus_encoding_stream.hpp
```diff
@@ -0,0 +1,143 @@
...
+      return data_content;
+    }
+
+   private:
+    boost::dynamic_bitset<uint8_t> content_{}; /**< RLE+ encoded content */
+
+    /**
+     * @brief Write RLE+ header
+     */
+    void initContent() {
```
Alexey-N-Chernyshov: Move non-template functions to .cpp

core/codec/rle/rle_plus_encoding_stream.hpp
```diff
@@ -0,0 +1,143 @@
...
+      content_.push_back(false);
+    }
+
+    /**
+     * @brief Write RLE+ small block
+     * @tparam T - type of block value
+     * @param block - value to write
+     */
+    template <typename T>
+    void pushSmallBlock(const T block) {
```
Alexey-N-Chernyshov: const T &block
I'm not sure for that.

core/codec/rle/rle_plus.hpp
sergkaprovich: My opinion: we don't need to use a `gsl::span` reference, because this type is pretty small (just pointer to start and sequence length) and copy operations are very cheap. 
I think, the closest notion of `gsl::span` is `std::string_view` - this type also recommended to be passed by value instead reference.

core/codec/rle/rle_plus_decoding_stream.hpp
sergkaprovich: Google C++ Style Guide doesn't restrict using output parameters [#Output_Parameters](https://google.github.io/styleguide/cppguide.html#Output_Parameters)
If we prefer use return values here, we will get code like this:
```
// Method signature:
std::pair<T, std::set<T>> decodeSingleBlock(const T& current_value, const std::set<T> &output);

auto result = decodeSingleBlock(value, output);
value = result.first;
value = result.second;
```
Do we really need to choose return values here instead output parameters?

core/codec/rle/rle_plus_decoding_stream.hpp
sergkaprovich: See above

core/codec/rle/rle_plus_decoding_stream.hpp
sergkaprovich: See above

core/codec/rle/rle_plus_encoding_stream.hpp
sergkaprovich: There are just 6 lines of code in the two non-template functions, I think we don't need additional source code file for this.

core/codec/rle/rle_plus_encoding_stream.hpp
sergkaprovich: Reference is a constant pointer with automatic redirection, so on x64 system `sizeof(T&) = 8`. We supports uint8_t...uint64_t, so in the case of using a value parameter we can spar several bytes, but in the case of using reference parameter we will always spend 8 bytes.

core/codec/rle/rle_plus_encoding_stream.hpp
sergkaprovich: See below

core/codec/rle/rle_plus_decoding_stream.hpp
masterjedy: We'd better had consistent code except it is absolutely necessary.

sergkaprovich: Using return values here instead output parameters will lead to unnecessary copy operations: in first case we will create std::set<T> and next append it to the result, but in the second case we will directly emplace new values to the result.
Maybe, we have another way to implement this actions without extra copy operations and output parameters?

kamilsa: Prefer return values instead of adding output parameters. 
There is no overhead in returning new set from this function. RVO will help to avoid redundant copy. All compilers do that

core/codec/rle/rle_plus_encoding_stream.hpp
kamilsa: And they will be copied and compile as many times as many includes we have for this file. I would prefer to create separate cpp, to avoid this

core/codec/rle/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+add_library(rle_plus_codec
+    rle_plus_errors.cpp
+    )
+
+target_link_libraries(rle_plus_codec
+    Boost::boost
+    outcome
+)
```
kamilsa:
```suggestion
    )
```

core/codec/rle/rle_plus_decoding_stream.hpp
sergkaprovich: Most likely I explained it unclear, please, look here again: it is not a RVO! 

In case of using return values:
1. We need to create new std::set and return it from function
2. Then we need to append this set to the result <- here is overhead copy operation

In case of using output parameters:
1. Emplace new values to the resulting set without additional copy operations

PR #26, opened 2019-12-12
add Address generation from private keys and validation

Alexey-N-Chernyshov:
### Description of the Change
Add construction from public keys.
Add validation.

PR #30, opened 2019-12-16
Actor

turuslan:
Actor interface, constants, cbor de/serialization

https://filecoin-project.github.io/specs/#systems__filecoin_vm__actor


cmake/Hunter/config.cmake
```diff
@@ -30,7 +30,7 @@ hunter_config(
 )
 
 hunter_config(libp2p
-    URL https://github.com/soramitsu/libp2p/archive/00ec26288abe3c70c3fab4b0529e02302a77fd20.zip
-    SHA1 b818125aa885a877d1f51836b8f632689b25766b
+    URL https://github.com/soramitsu/libp2p/archive/1e874e7eefb45b0b47e00aa73334411b031e385b.zip
```
ekovalev: Just a heads-up: this will conflict with https://github.com/filecoin-project/cpp-filecoin/pull/26/files (depending which one gets merged first)

core/actor/actor.cpp
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+#include "actor/actor.hpp"
+
+#include <string>
+#include <vector>
+
+#include <libp2p/crypto/sha/sha256.hpp>
+
+namespace fc::actor {
```
ekovalev: fc::vm::actor maybe? Actors don't make sense outside of VM

core/actor/actor.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include "primitives/big_int.hpp"
+
+namespace fc::actor {
+  using primitives::BigInt;
+  using libp2p::multi::ContentIdentifier;
+
+  struct Actor {
+    ContentIdentifier code;
+    ContentIdentifier head;
```
ekovalev: Are there any benefits in making the Actor's public fields names differ from their respective counterparts in the spec https://filecoin-project.github.io/specs/#systems__filecoin_vm__actor ?

turuslan: Names taken from lotus https://github.com/filecoin-project/lotus/blob/master/chain/types/actor.go#L11

core/vm/actor/actor.cpp
```diff
@@ -0,0 +1,50 @@
...
+  ContentIdentifier makeRawIdentityCid(const std::string &str) {
+    std::vector<uint8_t> bytes{str.begin(), str.end()};
+    return {ContentIdentifier::Version::V1,
+            libp2p::multi::MulticodecType::Code::RAW,
+            libp2p::multi::Multihash::create(libp2p::multi::HashType::identity,
+                                             bytes)
+                .value()};
+  }
+
+  ContentIdentifier kAccountCodeCid = makeRawIdentityCid("fil/1/account");
```
Harrm: Should these not be constants?

core/vm/actor/actor.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include <libp2p/multi/content_identifier_codec.hpp>
+
+#include "primitives/big_int.hpp"
+
+namespace fc::vm::actor {
+  using primitives::BigInt;
+  using libp2p::multi::ContentIdentifier;
+
+  struct Actor {
```
Harrm: Please comment.

core/vm/actor/actor.hpp
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  template <class Stream,
+            typename = std::enable_if_t<Stream::is_cbor_decoder_stream>>
+  Stream &operator>>(Stream &&s, Actor &actor) {
+    s.list() >> actor.code >> actor.head >> actor.nonce >> actor.balance;
+    return s;
+  }
+
+  bool isBuiltinActor(const ContentIdentifier &code);
```
Harrm: And these too. If the comments would not be too obvious.

core/vm/actor/actor.cpp
turuslan: `ContentIdentifier` initialization is not constant, for example `Multihash::create` uses outcome, so we cannot use `contstexpr`

PR #32, opened 2019-12-17
IPFS datastore

masterjedy:

### Description of the Change
IPFS datastore implementation based on Leveldb database.

### Benefits
IPFS datastore implemented

### Possible Drawbacks 
None

### DIFF
--- a/test/core/storage/ipfs/datastore_integration_test.cpp
+++ b/test/core/storage/ipfs/datastore_integration_test.cpp
@@ -29,7 +29,10 @@ struct DatastoreIntegrationTest : public ::testing::Test {
   static constexpr size_t kDefaultBufferLength = 32;
 
   static boost::filesystem::path makeTempPath() {
-    return boost::filesystem::temp_directory_path();
+    boost::filesystem::path global_temp_dir =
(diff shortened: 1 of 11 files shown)


test/core/storage/ipfs/datastore_integration_test.cpp
```diff
@@ -0,0 +1,146 @@
...
+using fc::storage::LevelDB;
+using fc::storage::ipfs::LeveldbDatastore;
+using libp2p::crypto::random::BoostRandomGenerator;
+using libp2p::crypto::random::CSPRNG;
+using libp2p::multi::ContentIdentifier;
+using libp2p::multi::HashType;
+using libp2p::multi::MulticodecType;
+using libp2p::multi::Multihash;
+
+struct DatastoreIntegrationTest : public ::testing::Test {
```
Alexey-N-Chernyshov: I suggest you to reuse `test::BaseFS_Test`. It creates tmp dir and cleans up.

masterjedy: Fixed temp dir creation and removal. I don't think I should use that base class, because it contains redundant data and profit is insignificant.

PR #34, opened 2019-12-17
Feature/keystore

Alexey-N-Chernyshov:
### Description of the Change
Keystore and in-memory and filesystem implementations.
Some small improvements.

### Benefits
Keystore is implemented.

### Possible Drawbacks 
no


core/storage/keystore/keystore.cpp
```diff
@@ -0,0 +1,96 @@
...
+using fc::storage::keystore::KeyStoreError;
+
+KeyStore::KeyStore(std::shared_ptr<BlsProvider> blsProvider,
+                   std::shared_ptr<Secp256k1Provider> secp256K1Provider,
+                   std::shared_ptr<AddressVerifier> addressVerifier)
+    : bls_provider_(std::move(blsProvider)),
+      secp256k1_provider_(std::move(secp256K1Provider)),
+      address_verifier_(std::move(addressVerifier)) {}
+
+fc::outcome::result<bool> KeyStore::CheckAddress(
```
turuslan: `boost::get` [throws](https://www.boost.org/doc/libs/1_55_0/doc/html/boost/get_idp140956568.html), use `visit_in_place` [(example)](https://github.com/filecoin-project/cpp-filecoin/blob/master/core/primitives/address.cpp#L29) or catch exception

core/storage/keystore/keystore.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+fc::outcome::result<bool> KeyStore::Verify(
+    const Address &address,
+    gsl::span<const uint8_t> data,
+    const TSignature &signature) noexcept {
+  OUTCOME_TRY(private_key, Get(address));
+  OUTCOME_TRY(valid, CheckAddress(address, private_key));
+  if (!valid) return KeyStoreError::WRONG_ADDRESS;
+
+  try {
```
turuslan: Mehtods called here use `outcome`, why do we need try-catch?

If this catches `boost::get(signature)` exceptions, then check signature type using `visit_in_place`.

core/storage/keystore/keystore.cpp
Alexey-N-Chernyshov: `visit_in_place` is unable in current implementation. Added todos.

core/storage/keystore/keystore.cpp
Alexey-N-Chernyshov: visit_in_place is unable in current implementation. Added todos.

core/storage/keystore/impl/in_memory/in_memory_keystore.cpp
```diff
@@ -0,0 +1,69 @@
...
+    std::shared_ptr<BlsProvider> blsProvider,
+    std::shared_ptr<Secp256k1Provider> secp256K1Provider,
+    std::shared_ptr<AddressVerifier> addressVerifier)
+    : KeyStore(std::move(blsProvider), std::move(secp256K1Provider),
+               std::move(addressVerifier)) {}
+
+fc::outcome::result<bool> InMemoryKeyStore::Has(
+    const Address &address) noexcept {
+  try {
+    return storage_.find(address) != storage_.end();
```
turuslan: `map.find` does not throw, remove try-catch

core/storage/keystore/impl/filesystem/filesystem_keystore.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+fc::outcome::result<void> FileSystemKeyStore::Put(
+    Address address, typename KeyStore::TPrivateKey key) noexcept {
+  OUTCOME_TRY(exists, Has(address));
+  if (exists) return KeyStoreError::ALREADY_EXISTS;
+  OUTCOME_TRY(path, AddressToPath(address));
+  OUTCOME_TRY(file, filestore_->create(path));
+
+  // TODO(a.chernyshov): use visit_in_place
```
turuslan: use `CheckAddress`

test/core/storage/keystore/filesystem_keystore_test.cpp
```diff
@@ -0,0 +1,318 @@
...
+/**
+ * @given Keystore with bls private key
+ * @when Verify() called with correct signature
+ * @then true returned
+ */
+TEST_F(FileSystemKeyStoreTest, VerifyCorrectBls) {
+  EXPECT_OUTCOME_TRUE_1(ks->Put(bls_address_, bls_keypair_->private_key));
+
+  auto signature = bls_provider_->sign(data_, bls_keypair_->private_key);
+  ASSERT_TRUE(signature);
```
turuslan: use `EXPECT_OUTCOME_TRUE(signature, bls_provider_->sign(...))`

test/core/storage/keystore/filesystem_keystore_test.cpp
```diff
@@ -0,0 +1,318 @@
...
+/**
+ * @given Kestore with secp256k1 private key
+ * @when Verify() called with invalid signature
+ * @then false returned
+ */
+TEST_F(FileSystemKeyStoreTest, InvalidSecp256k1Signature) {
+  EXPECT_OUTCOME_TRUE_1(
+      ks->Put(secp256k1_address_, secp256k1_keypair_->private_key));
+  Secp256k1Signature invalid_signature;
+  auto res = ks->Verify(secp256k1_address_, data_, invalid_signature);
```
turuslan: use `EXPECT_OUTCOME_ERROR(CryptoProviderError::SIGNATURE_VERIFICATION_FAILED, ks->Verify(...))`

test/core/storage/keystore/filesystem_keystore_test.cpp
```diff
@@ -0,0 +1,318 @@
...
+  ASSERT_FALSE(res);
+  ASSERT_EQ(KeyStoreError::NOT_FOUND, res.error());
+}
+
+/**
+ * @given Keystore with bls private key
+ * @when Verify() called with wrong signature
+ * @then false returned
+ */
+TEST_F(FileSystemKeyStoreTest, VerifyWrongBls) {
```
turuslan: why `Verify` behavior for secp256k1 and bls is inconsistent?
for invalid signature VerifyWrongBls test expects `outcome::success(false)` and InvalidSecp256k1Signature test expects `outcome::failure(CryptoProviderError::SIGNATURE_VERIFICATION_FAILED)`

test/core/storage/keystore/filesystem_keystore_test.cpp
```diff
@@ -0,0 +1,318 @@
...
+    address_verifier_ = std::make_shared<AddressVerifierImpl>();
+
+    ks = std::make_shared<FileSystemKeyStore>(
+        TEST_PATH, bls_provider_, secp256k1_provider_, address_verifier_);
+  }
+
+ protected:
+  static bool VectorContains(const std::vector<Address> &vector,
+                             Address value) {
+    return std::find(vector.begin(), vector.end(), value) != vector.end();
```
turuslan: `vector.find(value) != vector.end()`

core/storage/filestore/file.hpp
```diff
@@ -48,17 +48,17 @@ namespace fc::storage::filestore {
...
      * @param offset start position to write to
      * @param buffer pointer to the first element of array to be written
      * @return number of bytes written
      */
-    virtual fc::outcome::result<size_t> write(size_t offset,
-                                              const gsl::span<const uint8_t> &buffer) noexcept = 0;
+    virtual fc::outcome::result<size_t> write(
+        size_t offset, gsl::span<const uint8_t> buffer) noexcept = 0;
 
     /**
```
ekovalev: What's the rationale behind passing a span by value here?

core/storage/keystore/keystore.hpp
```diff
@@ -0,0 +1,118 @@
...
+     * @param address key identifier
+     */
+    virtual outcome::result<bool> Has(const Address &address) noexcept = 0;
+
+    /**
+     * @brief stores a key in the Keystore
+     * @param address - key idnetifier
+     * @param key
+     */
+    virtual outcome::result<void> Put(Address address,
```
ekovalev: Why return a variant here, not just KeyStoreError?

turuslan: It's common practice for our projects to return outcome::result to handle error-or-result.
For consistency methods that don't return value but may return error should return outcome::result<void>.

core/storage/filestore/file.hpp
turuslan: There is a convention in our projects to pass gsl::span by value rather than by reference.
gsl::span simply wraps two pointers, so it's cheap to copy.

test/core/storage/keystore/filesystem_keystore_test.cpp
Alexey-N-Chernyshov: Because any bls signature is valid. And test `VerifyWrongBls` verifies wrong signature. In case of secp256k1 signature with zeroes is invalid and signature algorithm fails in test `InvalidSecp256k1Signature`. Valid but wrong secp256k1 is tested in `WrongSecp256k1Signature`.

core/storage/filestore/file.hpp
Alexey-N-Chernyshov: It was my fault.

core/storage/keystore/keystore.hpp
Alexey-N-Chernyshov: To use wrappers OUTCOME_TRY....

core/storage/keystore/keystore.cpp
```diff
@@ -0,0 +1,97 @@
...
+    const Address &address,
+    gsl::span<const uint8_t> data,
+    const TSignature &signature) noexcept {
+  OUTCOME_TRY(private_key, Get(address));
+  OUTCOME_TRY(valid, CheckAddress(address, private_key));
+  if (!valid) return KeyStoreError::WRONG_ADDRESS;
+
+  return visit_in_place(
+      signature,
+      [this, address, data, private_key](
```
ekovalev: Do you need to capture "this"? Don't see it being used explicitly inside the lambda body

Alexey-N-Chernyshov: yes, for this-> bls_provider_

PR #39, opened 2019-12-19
Feature/repository

Alexey-N-Chernyshov:
### Description of the Change
Interface of Repository according to https://github.com/ipfs/specs/blob/master/REPO.md

### Benefits
Interface for in-memory and fs implementations.

### Usage Examples or Tests *[optional]*
In-memory and fs implementations will be later.


core/storage/repository/impl/repository_error.cpp
```diff
@@ -0,0 +1,19 @@
...
+  using fc::storage::repository::RepositoryError;
+
+  switch (e) {
+    case RepositoryError::WRONG_VERSION:
+      return "RepositoryError: wrong version";
+    case RepositoryError::UNKNOWN:
+      break;
+  }
+
+  return "unknown error";
```
artyom-yurin:
```suggestion
  return "RepositoryError: unknown error";
```

artyom-yurin:
```suggestion
  return "RepositoryError: unknown error";
```

core/storage/repository/repository_error.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+namespace fc::storage::repository {
+
+  /**
+   * @brief Type of errors returned by Repository
+   */
+  enum class RepositoryError {
+    WRONG_VERSION = 1,
+
+    UNKNOWN = 1000
```
masterjedy: don't need this assignment

PR #43, opened 2019-12-22
Cbor refactor

turuslan:
cbor: return outcome from encode, replace vector arg with gsl span, make encoder from bytes.
big_int: rvalue.
cbor_test: kDummyCid, EXPECT_OUTCOME_EQ.

### DIFF
--- a/core/codec/cbor/cbor.hpp
+++ b/core/codec/cbor/cbor.hpp
@@ -18,10 +18,14 @@ namespace fc::codec::cbor {
    * @return encoded data
    */
   template <typename T>
-  std::vector<uint8_t> encode(const T &arg) {
-    CborEncodeStream encoder;
-    encoder << arg;
-    return encoder.data();
+  outcome::result<std::vector<uint8_t>> encode(const T &arg) {
+    try {
+      CborEncodeStream encoder;
+      encoder << arg;
+      return encoder.data();
+    } catch (std::system_error &e) {
+      return outcome::failure(e.code());
+    }
   }
 
   /**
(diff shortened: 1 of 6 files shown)


core/codec/cbor/cbor.hpp
```diff
@@ -32,7 +36,7 @@ namespace fc::codec::cbor {
    * @see cbor_errors.hpp for possible error cases
    */
   template <typename T>
-  outcome::result<T> decode(const std::vector<uint8_t> &input) {
+  outcome::result<T> decode(gsl::span<const uint8_t> input) {
```
ekovalev: How useful is it? Because now the signature of ```decode``` is not anymore an inverse of ```encode```:  ```std::vector<uint8_t> -> T | error``` (even though the isomorphic property ```decode(encode(a)) == a``` will still hold).

masterjedy: span is better

core/codec/cbor/cbor.hpp
```diff
@@ -18,10 +18,14 @@ namespace fc::codec::cbor {
    * @return encoded data
    */
   template <typename T>
-  std::vector<uint8_t> encode(const T &arg) {
-    CborEncodeStream encoder;
-    encoder << arg;
-    return encoder.data();
+  outcome::result<std::vector<uint8_t>> encode(const T &arg) {
```
masterjedy: In order to provide this operator with ability to work with multiple arguments you can consult with scale implementation [here](https://github.com/soramitsu/kagome/blob/526ccc4766bbd8dd1b660be10e33ff730c00e710/core/scale/scale.hpp#L26)

turuslan: Filecoin uses IPLD, where each CBOR encoded object must have single root element.
I think it's better to leave it with single argument.

PR #44, opened 2019-12-23
HAMT

turuslan:


core/common/which.hpp
```diff
@@ -0,0 +1,38 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef CPP_FILECOIN_CORE_COMMON_WHICH_HPP
+#define CPP_FILECOIN_CORE_COMMON_WHICH_HPP
+
+#include <boost/variant.hpp>
+
+namespace fc::common {
+  template<typename T, typename T2>
```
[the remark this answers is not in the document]
Harrm: And, by the way, can you please explain why variant.type() can't be used for this purpose?

core/storage/hamt/hamt.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+  // assuming 8-bit indices
+  auto keyToIndices(const std::string &key, int n = -1) {
+    std::vector<uint8_t> key_bytes;
+    for (auto c : key) {
+      key_bytes.push_back(c);
+    }
+    auto hash = crypto::murmur::hash(key_bytes);
+    std::vector<size_t> result;
+    for (auto i = n == -1 ? hash.begin() : hash.end() - n + 1; i != hash.end();
```
Harrm: Please take the initialization of i outside with an ordinary if. The way it is now is hard to read.

core/storage/hamt/hamt.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+  using CID = libp2p::multi::ContentIdentifier;
+  using fc::primitives::UBigInt;
+  using Value = ipfs::IpfsDatastore::Value;
+
+  extern const CID kDummyCid;
+
+  constexpr size_t kLeafMax = 3;
+
+  /** Hamt node interface */
```
Harrm: I don't get this comment quite clearly. Why is it called an interface if it's a structure?

core/storage/hamt/hamt.hpp
```diff
@@ -0,0 +1,136 @@
...
+          l_pair >> key;
+          leaf.emplace(key, l_pair.raw());
+        }
+        node.items.emplace_back(leaf);
+      }
+    }
+    return s;
+  }
+
+  /** Hamt interface */
```
Harrm: This one does not seem to be an interface, either. Also I think it would be great if a short description or at least a link on what HAMT is was present here.

core/storage/ipfs/impl/datastore_leveldb.cpp
```diff
@@ -41,6 +41,7 @@ namespace fc::storage::ipfs {
   }
 
   outcome::result<void> LeveldbDatastore::set(const CID &key, Value value) {
+    // maybe check value hash matches cid
```
Harrm: As this was discussed on a meeting, just a reminder to remove this comment. Or make a TODO.

core/storage/ipfs/impl/in_memory_datastore.cpp
```diff
@@ -13,9 +13,8 @@ fc::outcome::result<bool> InMemoryDatastore::contains(const CID &key) const {
 }
 
 fc::outcome::result<void> InMemoryDatastore::set(const CID &key, Value value) {
-  auto res = storage_.emplace(key, value);
-  if (!res.second) return IpfsDatastoreError::CANNOT_STORE;
-
+  // maybe check value hash matches cid
```
Harrm: And here.

test/core/storage/hamt/hamt_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+}
+
+class HamtTest : public ::testing::Test {
+ public:
+  std::shared_ptr<fc::storage::ipfs::IpfsDatastore> store_{std::make_shared<fc::storage::ipfs::InMemoryDatastore>()}; 
+  std::shared_ptr<Node> root_{std::make_shared<Node>()};
+  Hamt hamt_{store_, root_};
+};
+
+TEST_F(HamtTest, NodeCbor) {
```
Harrm: Please supply all tests with given/when/then comments.

core/storage/hamt/hamt.cpp
```diff
@@ -0,0 +1,247 @@
...
+  Hamt::Hamt(std::shared_ptr<ipfs::IpfsDatastore> store, Node::Ptr root)
+      : store_(std::move(store)), root_(root) {}
+
+  Hamt::Hamt(std::shared_ptr<ipfs::IpfsDatastore> store, const CID &root)
+      : store_(std::move(store)), root_(root) {}
+
+  outcome::result<void> Hamt::set(const std::string &key,
+                                  gsl::span<const uint8_t> value) {
+    OUTCOME_TRY(loadItem(root_));
+    return set(*boost::get<Node::Ptr>(root_), keyToIndices(key), key, value);
```
Harrm: Should here not be a check whether root_ contains Node::Ptr?

core/storage/hamt/hamt.cpp
```diff
@@ -0,0 +1,247 @@
...
+  }
+
+  outcome::result<Value> Hamt::get(const std::string &key) {
+    OUTCOME_TRY(loadItem(root_));
+    auto node = boost::get<Node::Ptr>(root_);
```
Harrm: And also here.

core/storage/hamt/hamt.cpp
```diff
@@ -0,0 +1,247 @@
...
+      return "Max depth exceeded";
+    default:
+      return "Unknown error";
+  }
+}
+
+namespace fc::storage::hamt {
+  using fc::common::which;
+
+  const CID kDummyCid({}, {}, libp2p::multi::Multihash::create({}, {}).value());
```
masterjedy: don't think it is a good name for an instance in hpp.

core/storage/hamt/hamt.cpp
masterjedy: pls simplify

core/storage/hamt/hamt.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+  // assuming 8-bit indices
+  auto keyToIndices(const std::string &key, int n = -1) {
+    std::vector<uint8_t> key_bytes;
+    for (auto c : key) {
+      key_bytes.push_back(c);
+    }
+    auto hash = crypto::murmur::hash(key_bytes);
+    std::vector<size_t> result;
```
masterjedy: use result.reserve()

core/storage/hamt/hamt.cpp
```diff
@@ -0,0 +1,247 @@
...
+}
+
+namespace fc::storage::hamt {
+  using fc::common::which;
+
+  const CID kDummyCid({}, {}, libp2p::multi::Multihash::create({}, {}).value());
+
+  // assuming 8-bit indices
+  auto keyToIndices(const std::string &key, int n = -1) {
+    std::vector<uint8_t> key_bytes;
```
masterjedy: use key_bytes.reserve(key.length());

core/storage/hamt/hamt.cpp
```diff
@@ -0,0 +1,247 @@
...
+    for (auto c : key) {
```
masterjedy: probably you can use std::copy()

core/storage/hamt/hamt.cpp
masterjedy: Code below seems to work. Ensure that argument of subspan is correct
```std::vector<uint8_t> key_bytes(key.begin(), key.end());
    auto hash = crypto::murmur::hash(key_bytes);
    auto span = gsl::make_span(hash).subspan(key.size() - n);
    return std::vector<size_t>(span.begin(), span.end());
```
Looks prettier.

core/storage/hamt/hamt.cpp
```diff
@@ -0,0 +1,247 @@
...
+    return CID(CID::Version::V1, libp2p::multi::MulticodecType::DAG_CBOR, hash);
+  }
+
+  outcome::result<CID> Node::cid() const {
+    OUTCOME_TRY(encoded, codec::cbor::encode(*this));
+    return cid(encoded);
+  }
+
+  Hamt::Hamt(std::shared_ptr<ipfs::IpfsDatastore> store, Node::Ptr root)
+      : store_(std::move(store)), root_(root) {}
```
masterjedy:
```suggestion
      : store_(std::move(store)), root_(std::move(root)) {}
```

core/storage/hamt/hamt.cpp
masterjedy: Actually it doesn't work, I checked. Something really weird happens here. 
In tests sometimes index goes out of bound. Without spans it can silently pass becoming a floating bug. Please recheck your implementation and tests and rewrite based on the code I specified.

core/storage/hamt/hamt.cpp
```diff
@@ -0,0 +1,247 @@
...
+      for (auto &pair : leaf) {
+        auto indices2 = keyToIndices(pair.first, indices.size());
+        OUTCOME_TRY(set(*child, indices2, pair.first, pair.second));
+      }
+      item = child;
+    }
+    return outcome::success();
+  }
+
+  outcome::result<void> Hamt::remove(Node &node,
```
masterjedy: method is too complex, please split it or simplify

core/storage/hamt/hamt.cpp
```diff
@@ -0,0 +1,247 @@
...
+      item = cid;
+    }
+    return outcome::success();
+  }
+
+  outcome::result<void> Hamt::loadItem(Node::Item &item) {
+    if (which<CID>(item)) {
+      OUTCOME_TRY(child_bytes, store_->get(boost::get<CID>(item)));
+      OUTCOME_TRY(child, codec::cbor::decode<Node>(child_bytes));
+      item = std::make_shared<Node>(child);
```
masterjedy:
```suggestion
      item = std::make_shared<Node>(std::move(child));
```

core/storage/hamt/hamt.cpp
turuslan: `Hamt::loadItem(Node::Item)` ensures that node is either `Node::Ptr` or `Node::Leaf`.
`root_` is never `Node::Leaf`.

core/common/which.hpp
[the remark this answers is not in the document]
turuslan: `boost::variant::type()` requires RTTI, which may cause problems.
`boost::variant::which()` uses just integer index to template type argument.

core/storage/hamt/hamt.cpp
turuslan: Thanks, didn't know about this posibility.

core/storage/hamt/hamt.cpp
```diff
@@ -31,17 +30,10 @@ namespace fc::storage::hamt {
...
-    }
+    std::vector<uint8_t> key_bytes(key.begin(), key.end());
     auto hash = crypto::murmur::hash(key_bytes);
-    std::vector<size_t> result;
-    for (auto i = n == -1 ? hash.begin() : hash.end() - n + 1; i != hash.end();
-         ++i) {
-      result.push_back(*i);
-    }
-    return result;
+    return std::vector<size_t>(n == -1 ? hash.begin() : hash.end() - n + 1,
```
masterjedy: return statement is too complex

test/core/storage/hamt/hamt_test.cpp
```diff
@@ -43,6 +43,7 @@ class HamtTest : public ::testing::Test {
   Hamt hamt_{store_, root_};
 };
 
+/** Hamt node CBOR encoding and decoding, correct CID */
```
masterjedy: according to soramitsu's practice each test should have comments it form
```
/**
 * @given ...
 * @when ...
 * @then ...
*/
```

PR #45, opened 2019-12-24
Feature/in memory repository

Alexey-N-Chernyshov:
### Description of the Change
See [interface PR](https://github.com/filecoin-project/cpp-filecoin/pull/39) first.
In-memory implementation of Repository.


core/storage/repository/impl/in_memory_repository.hpp
```diff
@@ -0,0 +1,40 @@
...
+  static constexpr Version kInMemoryRepositoryVersion = 1;
+
+  /**
+   * @brief In-memory implementation of Repository
+   */
+  class InMemoryRepository : public Repository {
+   public:
+    InMemoryRepository();
+
+    ~InMemoryRepository() override = default;
```
sergkaprovich: I think, that this default override destructor doesn't affect behaviour of this code and can be omitted
[https://stackoverflow.com/questions/40999791/default-override-of-virtual-destructor/41001264](https://stackoverflow.com/questions/40999791/default-override-of-virtual-destructor/41001264)

PR #50, opened 2019-12-25
introduce vm ExitCode

Alexey-N-Chernyshov:

### Description of the Change
Introduce VM Exit Codes according to the [specification](https://filecoin-project.github.io/specs/#systems__filecoin_vm__runtime__exitcode).
Add to_integer for enum constants.

### Benefits
Exit Codes for VM Runtime.


core/vm/exit_code/exit_code.hpp
```diff
@@ -0,0 +1,182 @@
...
+    // does not satisfy its preconditions.
+    kRuntimeAPIError,
+
+    // Returned when an actor method invocation calls rt.Assert with a false
+    // condition.
+    kRuntimeAssertFailure,
+
+    // Returned when an actor method's Send call has returned with a failure
+    // error code (and the Send call did not specify to ignore errors).
+    kMethodSubcallError
```
igor-egorov:
```suggestion
    kMethodSubcallError,
```

core/vm/exit_code/exit_code.hpp
```diff
@@ -0,0 +1,182 @@
...
+     * @return true
+     */
+    bool operator==(const Success &other) const;
+  };
+
+  /**
+   * @brief System error codes
+   */
+  enum class SystemError {
+    // Represents a failure to find an actor.
```
igor-egorov: Since those are docstrings, let's use /// or /** */ for multiline strings here and below.

core/vm/exit_code/exit_code.hpp
```diff
@@ -0,0 +1,182 @@
...
+    kActorNotFound = 1,
+
+    // Represents a failure to find the code for a particular actor in the VM
+    // registry.
+    kActorCodeNotFound,
+
+    // Represents a failure to find a method in an actor
```
igor-egorov: How do you decide whether to put the trailing dot? )

core/vm/exit_code/exit_code.hpp
```diff
@@ -0,0 +1,182 @@
...
+    // Returned when an actor method's Send call has returned with a failure
+    // error code (and the Send call did not specify to ignore errors).
+    kMethodSubcallError
+  };
+
+  /**
+   * @brief User defined error codes
+   */
+  enum class UserDefinedError {
+    kInsufficientFundsUser = 1,
```
igor-egorov: Exhaustingly documented enum constants above and nothing here 🤔

core/vm/exit_code/exit_code.hpp
```diff
@@ -0,0 +1,182 @@
...
+    kDeadlineExceeded,
+    kInsufficientPledgeCollateral,
+  };
+
+  /**
+   * @brief Virtual machine exit code
+   */
+  class ExitCode {
+   public:
+    using TValue = boost::variant<Success, SystemError, UserDefinedError>;
```
igor-egorov: It does not look like the right solution for error handling. Better to have a single type (or a common parent) for the error case. The reason is - when TValue is received, then you have to employ two visitors instead of a single for error processing. The current approach is less maintainable and performant at runtime.

core/common/enum.hpp
```diff
@@ -0,0 +1,39 @@
...
+   *   @code
+   *   int v = as_integer(Enumerator::eleven); // v == 11
+   *   @nocode
+   *
+   * @tparam Enumeration - enum type
+   * @param value - to convert
+   * @return integer value of enum
+   */
+  template <typename Enumeration>
+  auto as_integer(Enumeration const value) ->
```
igor-egorov: Just a suggestion. Maybe it worth to rename to sound more like std::to_string for consistency purposes.
```suggestion
  auto to_int(Enumeration const value) ->
```

core/vm/exit_code/impl/exit_code.cpp
```diff
@@ -0,0 +1,96 @@
...
+    : exit_code_(std::move(exit_code)),
+      error_message_(std::move(error_message)) {}
+
+RuntimeError::RuntimeError(ExitCode exit_code)
+    : exit_code_(std::move(exit_code)) {}
+
+std::string RuntimeError::toString() const {
+  std::stringstream res;
+  res << "Runtime error: '" << exit_code_.toString();
+  if (!error_message_.empty()) res << " (\"" << error_message_ << "\")";
```
igor-egorov: do we need extra quotes inside parenthesis?

core/vm/exit_code/exit_code.hpp
Alexey-N-Chernyshov: like in spec

igor-egorov:
Just fyi. I did not check the most of docstrings for typos and other issues

core/common/enum.hpp
```diff
@@ -0,0 +1,39 @@
...
+   *   @nocode
+   *   Can be converted as:
+   *   @code
+   *   int v = as_integer(Enumerator::eleven); // v == 11
+   *   @nocode
+   *
+   * @tparam Enumeration - enum type
+   * @param value - to convert
+   * @return integer value of enum
+   */
```
art-gor: do you really need such an excessive description for a mere static_cast?

core/vm/exit_code/exit_code.hpp
Alexey-N-Chernyshov: Actually no idea what is kInvalidSectorPacking or kInsufficientPledgeCollateral.

core/vm/exit_code/impl/exit_code.cpp
Alexey-N-Chernyshov: I would leave it according to the spec: `fmt.Sprintf(" (\"%v\")", x.ErrMsg)`

core/common/enum.hpp
Alexey-N-Chernyshov: The code is not self-explaining, so I left usage example.

core/vm/exit_code/exit_code.hpp
Alexey-N-Chernyshov: reworked

PR #51, opened 2019-12-25
refactor address

masterjedy:

### Description of the Change
Address refactored.

### Benefits
Address refactored.

### Possible Drawbacks 
None.


core/primitives/address/address_builder.hpp
```diff
@@ -6,11 +6,11 @@
 #ifndef CPP_FILECOIN_CORE_PRIMITIVES_ADDRESS_ADDRESS_BUILDER_HPP
 #define CPP_FILECOIN_CORE_PRIMITIVES_ADDRESS_ADDRESS_BUILDER_HPP
 
-#include <libp2p/crypto/secp256k1_types.hpp>
+#include "crypto/secp256k1_provider/secp256k1_provider.hpp"
 
 #include "common/outcome.hpp"
 #include "crypto/bls_provider/bls_types.hpp"
-#include "primitives/address.hpp"
+#include "address.hpp"
```
igor-egorov: I bet it is not clang-formatted (the order of includes)

core/primitives/address/impl/address_builder_impl.cpp
```diff
@@ -6,28 +6,31 @@
...
   using crypto::blake2b::blake2b_160;
+  using primitives::address::decode;
+  using primitives::address::encode;
+  using primitives::address::Protocol;
 
   outcome::result<Address> AddressBuilderImpl::makeFromSecp256k1PublicKey(
       Network network, const Sec256k1PublicKey &public_key) noexcept {
     OUTCOME_TRY(hash, blake2b_160(public_key));
-    std::vector<uint8_t> sec256k1_bytes{network, primitives::SECP256K1};
+    std::vector<uint8_t> sec256k1_bytes{network, Protocol::SECP256K1};
```
igor-egorov: not an issue, just to know. why Secp256k1 is all capitalized?

igor-egorov: BLS is ok since it is an acronym

Alexey-N-Chernyshov:
Also consider renaming `core/crypto/bls_provider' to 'core/crypto/bls'.

core/primitives/address/impl/address_builder_impl.cpp
masterjedy: it is also an acronym) anyway, if memory serves me right, we decided to use upper-case in constants. sometimes it applies to enums members, but it is not fully legal.

core/primitives/address/address_builder.hpp
masterjedy: fixed

PR #55, opened 2019-12-27
Refactor address_codec to render Address serialization consistent with Lotus

ekovalev:

### Description of the Change

Filecoin specification defines an Address as a product type of a network (mainnet or testnet) and a payload. However, existing Lotus implementation does not take the network part of the Address into account (furthermore, it is not even stored in the Address object). Subsequently, the network information is omitted during address serialization into a byte array and only appears in a human-readable string representation.

In order to be consistent with the way the addresses are passed around in Lotus testnet, the network byte is removed from the byte encoding of an address.

Apart from the above CBOR encoding is added for the Address which will be needed for some upcoming changed.

### Benefits

Consistency with Lotus network.

### Possible Drawbacks 

Potential discrepancy with the specification. This needs to be addressed separately.

### DIFF
--- a/core/primitives/address/impl/address_builder_impl.cpp
+++ b/core/primitives/address/impl/address_builder_impl.cpp
@@ -15,7 +15,6 @@ namespace fc::primitives::address {
   using BlsPublicKey = crypto::bls::PublicKey;
   using crypto::blake2b::blake2b_160;
   using primitives::address::BLSPublicKeyHash;
-  using primitives::address::decode;
   using primitives::address::Protocol;
   using primitives::address::Secp256k1PublicKeyHash;
 
--- a/core/primitives/address/impl/address_builder_impl.cpp
+++ b/core/primitives/address/impl/address_builder_impl.cpp
@@ -14,23 +14,22 @@ namespace fc::primitives::address {
   using Sec256k1PublicKey = libp2p::crypto::secp256k1::PublicKey;
   using BlsPublicKey = crypto::bls::PublicKey;
   using crypto::blake2b::blake2b_160;
+  using primitives::address::BLSPublicKeyHash;
   using primitives::address::decode;
-  using primitives::address::encode;
   using primitives::address::Protocol;
+  using primitives::address::Secp256k1PublicKeyHash;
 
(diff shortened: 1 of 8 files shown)


core/primitives/address/impl/address_builder_impl.cpp
```diff
@@ -21,14 +21,14 @@ namespace fc::primitives::address {
   outcome::result<Address> AddressBuilderImpl::makeFromSecp256k1PublicKey(
       Network network, const Sec256k1PublicKey &public_key) noexcept {
     OUTCOME_TRY(hash, blake2b_160(public_key));
-    std::vector<uint8_t> sec256k1_bytes{network, Protocol::SECP256K1};
```
turuslan: Can we avoid calling `decode`?
Something like `return Address{MAINNET, Secp256k1PublicKeyHash(hash)};`
Same for bls.

Also may add TODO for [FIL-118](https://soramitsu.atlassian.net/browse/FIL-118).

test/core/primitives/address/address_verifier_test.cpp
```diff
@@ -183,7 +183,7 @@ TEST_P(AddressVerifierParametrizedTest, GenerateSecp256k1Address) {
   EXPECT_OUTCOME_TRUE(
       res, address_verifier->verifySyntax(address, keypair.public_key));
   ASSERT_TRUE(res);
-  ASSERT_EQ(GetParam(), address.network);
+  ASSERT_EQ(Network::TESTNET, address.network);
```
Alexey-N-Chernyshov: Please, remove parametrized test as it is not used anymore.

Alexey-N-Chernyshov: I mean `INSTANTIATE_TEST_CASE_P()` and `TEST_P` -> `TEST_F`

core/primitives/address/impl/address_builder_impl.cpp
Alexey-N-Chernyshov: Secp256k1 and BLS Addresses are constructed in different ways. Secp256k1 uses hash of public key when bls just copies it. In my opinion code of creation shouldn't duplicate existing code even if it simple.

PR #58, opened 2019-12-31
FIL-72 Message component

ekovalev:

### Description of the Change

Adding the Message component (https://filecoin-project.github.io/specs/#systems__filecoin_vm__message).

### Benefits

Obvious.

### Possible Drawbacks 

Shouldn't be any. Need to watch out for potential inconsistencies with the existing implementation (Lotus) though to ensure different implementations are compatible with one another and can run in one network.


core/vm/message/message.hpp
```diff
@@ -0,0 +1,90 @@
...
+   */
+  struct UnsignedMessage {
+    Address to;
+    Address from;
+
+    uint64_t nonce;
+
+    BigInt value;
+
+    uint64_t method;
```
masterjedy: The order may be important to not get confused in case of serialization (I'm not sure if it is required).
Originally, the order is {to, from, nonce, value, gas price, gas limit, method, params}.

ekovalev: agree. The order of field members follows the spec while the order in serialized form matches Lotus implementation (which is a must for compatibility purpose)

core/vm/message/message_codec.hpp
```diff
@@ -0,0 +1,125 @@
...
+      typename = std::enable_if_t<
+          std::is_same_v<T,
+                         UnsignedMessage> || std::is_same_v<T, SignedMessage>>>
+  static std::vector<uint8_t> serialize(const T &m) {
+    CborEncodeStream encoder;
+    encoder << m;
+    return encoder.data();
+  }
+
+  /**
```
turuslan: Remove this and use `codec::cbor::decode<UnsignedMessage>(span)` directly.

core/vm/message/message_codec.hpp
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  template <class Stream,
+            typename = std::enable_if_t<Stream::is_cbor_decoder_stream>>
+  Stream &operator>>(Stream &s, SignedMessage &sm) {
+    s.list() >> sm.message >> sm.signature;
+    return s;
+  }
+
+  /**
```
turuslan: Remove this and use `codec::cbor::encode(message)` directly.
Note that it returns `outcome::result`.

core/vm/message/message.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+  /**
+   * @brief SignedMessage struct
+   */
+  struct SignedMessage {
+    UnsignedMessage message;
+    Signature signature;
+  };
+
+  extern const uint64_t kMessageMaxSize;
```
turuslan: Use `constexpr uint64_t kMessageMaxSize = 32 * 1024`, same for other constants.

core/vm/message/message_signer.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace fc::vm::message {
+
+  using PrivateKey = gsl::span<uint8_t>;
+  using PublicKey = gsl::span<uint8_t>;
+
+  /**
+   * Interface of an UnisgnedMessage Signer.
+   */
+  class MessageSigner {
```
Alexey-N-Chernyshov: According https://filecoin-project.github.io/specs/#system-requirements private keys must be secret, that's why we have KeyStore with `sign()` and `verify()` methods. I cannot find usage of MessageSigner, imho we should use KeyStore for signing.

core/vm/message/impl/message_signer_secp256k1.cpp
```diff
@@ -0,0 +1,61 @@
...
+    }
+    if (static_cast<uint64_t>(key.size())
+        != crypto::secp256k1::kPublicKeyLength) {
+      return outcome::failure(MessageError::INVALID_KEY_LENGTH);
+    }
+    Secp256k1PublicKey publicKey{};
+    std::copy_n(
+        key.begin(), crypto::secp256k1::kPublicKeyLength, publicKey.begin());
+
+    auto bytes = boost::get<Signature::Secp256k1Signature>(msg.signature.data);
```
Alexey-N-Chernyshov: it can be done with visit_in_place.

core/vm/message/message.hpp
```diff
@@ -0,0 +1,90 @@
...
+      using vector::vector;
+    };
+
+    struct BlsSignature : public crypto::bls::Signature {
+      using array::array;
+    };
+
+    enum Type : uint8_t { SECP256K1 = 0x1, BLS = 0x2 };
+
+    Type type;
```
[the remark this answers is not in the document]
Alexey-N-Chernyshov: Is it needed for cbor-encoding compatibility with lotus?

core/vm/message/message_util.cpp
```diff
@@ -0,0 +1,27 @@
...
+namespace fc::vm::message {
+
+  using common::getCidOf;
+
+  outcome::result<CID> cid(const UnsignedMessage &m) {
+    auto encoded = serialize<UnsignedMessage>(m);
+    return getCidOf(encoded);
+  }
+
+  outcome::result<CID> cid(const SignedMessage &sm) {
```
Alexey-N-Chernyshov: cid(bls_signed) == cid(unsigned)? And different for secp256k1? Is it correct?

ekovalev: yes, at least as long as the Lotus implementation can be trusted

core/vm/message/message_codec.hpp
ekovalev: That is the exact justification of this function to exist: for the sake of ease of composition with the ```sign``` method of the MessageSigner (```sign``` is pre-composed with ```serialize```) I needed a fucntion that would return a ```vector```, not ```outcome::result```). In this case none of the involved ```<<``` operators can throw so the entire machinery of the ```outcome::result``` is unnecessary.
However, since the Message signing mechanism is yet to be confirmed (see the concerns about private keys being exposed) I think this all still might change

core/vm/message/message.hpp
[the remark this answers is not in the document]
ekovalev: Briefly, yes - it's a copy of what they have in Lotus. The "type" byte is present in cbor encoding of the signature. Regarding whether or not we need to have it our Signature class... probably we could do without out just as well

masterjedy:
Lotus implementation has cbor encode and decode methods for Message and UnsignedMessage
[here](https://github.com/filecoin-project/lotus/blob/fda3811dcfcff4005a7cdbada2b1cc09ff8a75f9/chain/types/cbor_gen.go#L710). I suppose we will need those methods implemented in cpp-filecoin.

core/crypto/signature/signature.cpp
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "crypto/signature/signature.hpp"
+#include "common/outcome.hpp"
```
turuslan: Unnecessary include, signature.hpp already includes it

core/crypto/signature/signature.hpp
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * @brief Signature error codes
+   */
+  enum class SignatureError {
+    INVALID_SIGNATURE_LENGTH = 1,
+    WRONG_SIGNATURE_TYPE,
+    INVALID_KEY_LENGTH
+  };
+
+  // using Signature = boost::variant<BlsSignature, Secp256k1Signature>;
```
turuslan: Remove commented code

core/storage/keystore/keystore.hpp
```diff
@@ -77,8 +78,8 @@ namespace fc::storage::keystore {
      * @param data to sign
      * @return signature
      */
-    virtual outcome::result<TSignature> sign(const Address &address,
-                                             gsl::span<uint8_t> data) noexcept;
+    virtual outcome::result<Signature> sign(const Address &address, gsl::span<uint8_t> data) noexcept;
```
turuslan: gsl::span<**const** uint8_t>

core/vm/message/message_util.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * @brief SignedMessage size
+   */
+  outcome::result<uint64_t> size(const SignedMessage &sm);
+
+  /**
+   * @brief Message equality operator
+   */
+  inline bool operator==(const UnsignedMessage &lhs,
```
turuslan: `UnsignedMessage` cid equality is equivalent to `UnsignedMessage` fields equality.
Second should be more efficient as it does not encode and hash.
Write field-wise equality comparison.

test/core/primitives/address/address_codec_test.cpp
```diff
@@ -129,31 +129,43 @@ TEST_F(AddressCodecTest, EncodeToString) {
...
  * @then The outputs match expectations
  */
 TEST_F(AddressCodecTest, MarshalCbor) {
-  EXPECT_OUTCOME_TRUE_2(addr1, decode("0001"_unhex));
-  CborEncodeStream s1;
-  s1 << addr1;
-  EXPECT_EQ(s1.data(), "420001"_unhex);
+  EXPECT_OUTCOME_TRUE(addr1, decode("0001"_unhex));
+  EXPECT_OUTCOME_TRUE(cbor_encoded, fc::codec::cbor::encode<Address>(addr1));
+  EXPECT_EQ(cbor_encoded, "420001"_unhex);
```
turuslan: `EXPECT_OUTCOME_EQ(fc::codec::cbor::encode<Address>(addr1), "420001"_unhex)`

core/vm/message/message.hpp
```diff
@@ -0,0 +1,69 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef CPP_FILECOIN_CORE_VM_MESSAGE_HPP
+#define CPP_FILECOIN_CORE_VM_MESSAGE_HPP
+
+#include <boost/variant.hpp>
```
turuslan: Unnecessary includes <boost/variant.hpp>, "crypto/bls/bls_provider.hpp", "crypto/secp256k1/secp256k1_provider.hpp"

core/vm/message/impl/message_signer_impl.hpp
```diff
@@ -0,0 +1,43 @@
...
+namespace fc::vm::message {
+
+  using storage::keystore::KeyStore;
+
+  /**
+   * Class MessageSignerImpl.
+   */
+  class MessageSignerImpl : public MessageSigner {
+   public:
+    MessageSignerImpl() = delete;
```
Alexey-N-Chernyshov: I think there is no sense to delete default constructor explicitly since implicitly-default constructor is created only if no user-defined constructors are provided. But there is the user-defined constructor on the next line, so it is not the case.

core/vm/message/impl/message_signer_impl.hpp
```diff
@@ -0,0 +1,43 @@
...
+     * @copydoc MessageSigner::sign()
+     */
+    outcome::result<SignedMessage> sign(
+        const Address &address, const UnsignedMessage &msg) noexcept override;
+
+    /**
+     * @copydoc MessageSigner::verify()
+     */
+    outcome::result<UnsignedMessage> verify(
+        const Address &address, const SignedMessage &msg) noexcept override;
```
Alexey-N-Chernyshov:
```suggestion
        const Address &address, const SignedMessage &msg) const noexcept override;
```
It could be marked const

core/crypto/signature/signature.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+OUTCOME_CPP_DEFINE_CATEGORY(fc::crypto::signature, SignatureError, e) {
+  using fc::crypto::signature::SignatureError;
+  switch (e) {
+    case (SignatureError::INVALID_SIGNATURE_LENGTH):
+      return "Invalid signature length";
```
Alexey-N-Chernyshov:
```suggestion
      return "SignatureError: Invalid signature length";
```
Here and below. Just to be consistent.

PR #42, opened 2019-12-20
in-memory ipfs datastore implementation

Alexey-N-Chernyshov:
Signed-off-by: Alexey-N-Chernyshov <alexey.n.chernyshov@gmail.com>

In-memory IPFS Datastore
Add IPFS Datastore error
Rework IPFS tests

### Description of the Change

Introduce IPFS datastore in-memory implementation

PR #61, opened 2020-01-10
Tickets

masterjedy:

### Description of the Change
Feature tickets partially implemented.

### Benefits
Structures are implemented as well as necessary methods and we can go on with dependent tasks.

### Possible Drawbacks 
Tests are not ready yet.


test/core/primitives/ticket/ticket_test.cpp
```diff
@@ -0,0 +1,58 @@
...
+  auto &&t2 = generator.makeTicket();
+  ASSERT_FALSE(t1 == t2);
+}
+
+/**
+ * @given two different tickets t1, t2
+ * @when compare them using operator< (`less`)
+ * @then one and only one of t1 < t2 and t2 < t1 is true
+ */
+TEST_F(TicketTest, TwoTicketsLessSuccess) {
```
turuslan: Use pregenerated constants instead of runtime random generated.
Assertion is ambiguos, "less" implementation with wrong ordering will pass.

test/testutil/primitives/ticket_utils.hpp
```diff
@@ -0,0 +1,51 @@
...
+#include "primitives/ticket/epost_ticket.hpp"
+#include "primitives/ticket/ticket.hpp"
+
+#include <libp2p/crypto/random_generator/boost_generator.hpp>
+
+/**
+ * @class TicketGenerator provides methods for creating random tickets and
+ * related objects
+ */
+class TicketGenerator {
```
turuslan: See [comment for #42](https://github.com/filecoin-project/cpp-filecoin/pull/42#discussion_r360801059) for possible reasons to remove random generation from tests.

core/primitives/ticket/ticket.hpp
```diff
@@ -0,0 +1,53 @@
...
+   */
+  struct Ticket {
+    crypto::vrf::VRFProof bytes;
+  };
+
+  /**
+   * @brief compares 2 tickets
+   * @param lhs first ticket
+   * @param rhs second ticket
+   * @return true if equal false otherwise
```
Alexey-N-Chernyshov:
```suggestion
   * @return true if equal, false otherwise
```

test/testutil/primitives/ticket_utils.hpp
Alexey-N-Chernyshov: Agree with @turuslan.

masterjedy: I agree that we shouldn't rely only on random generated data, but such data is useful to cover cases not affordable otherwise. The argument about debug difficulties should be taken into account, I provide test with more debug information.

test/core/primitives/ticket/ticket_test.cpp
masterjedy: Agree that from such point of view this test has a weak point. I'm going to add a 'less' test as you described and rename this one, since its goal is to ensure, that the 'less' operation is antisymmetric.

test/core/primitives/ticket/ticket_test.cpp
```diff
@@ -33,26 +54,57 @@ TEST_F(TicketTest, SameTicketEqualSuccess) {
...
+  EXPECT_OUTCOME_TRUE_1(drawRandomness(t1, 1u));
+}
+
+/**
+ * @given predefined ticket, round value and lotus-drawn randomness
+ * @when call drawRandomness using given ticket and round value
+ * @then resulting randomness value is equal to lotus-calculated value
+ */
+TEST_F(TicketTest, DrawRandomnessLotusSuccess) {
+  EXPECT_OUTCOME_TRUE(randomness, drawRandomness(ticket_small, round_value));
```
turuslan: `EXPECT_OUTCOME_EQ(drawRandomness(ticket_small, round_value), randomness_value)`

masterjedy: I don't think it is better.

PR #63, opened 2020-01-13
Feature/code loading

turuslan:
### Description of the Change
Add `VMExitCode`.
Add `Invoker` that invokes builtin actor code.
Refactor `CronActor`.

### Benefits

Invoking builtin actor code.

### Possible Drawbacks 

None


.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         # 3 jobs in total
-        os: [ubuntu-18.04, macOS-10.14]
+        os: [ubuntu-18.04, macOS-latest]
```
masterjedy: why is it necessary?

core/vm/exit_code/exit_code.hpp
```diff
@@ -8,6 +8,16 @@
 
 #include <string>
 
+#include "common/outcome.hpp"
+
+namespace fc::vm {
+  enum class VMExitCode : uint8_t {};
```
masterjedy: empty enum? is it what you meant?

.github/workflows/ci.yml
turuslan: rebased wrong branch

core/vm/exit_code/exit_code.hpp
turuslan: Outcome errors use error enums.
This enum is empty because exit codes have different meaning across vm and different actors.

PR #67, opened 2020-01-15
Refactor CID

turuslan:
### Description of the Change

Replace `ContentIdentifier` (<libp2p/multi/content_identifier.hpp>) with `CID` ("core/common/cid.hpp").
Make CID default-constructible.

### Benefits

Default constructible CID.
Consistent usage of CID.

### Possible Drawbacks 

None.


core/codec/cbor/cbor_common.hpp
```diff
@@ -11,6 +11,7 @@
 #include <libp2p/multi/content_identifier_codec.hpp>
 
 #include "codec/cbor/cbor_errors.hpp"
+#include "common/cid.hpp"
```
artyom-yurin: Why is it here?

PR #77, opened 2020-01-29
IPFS MerkleDAG service

sergkaprovich:
### Description of the Change

IPFS MerkleDAG service provides functionality to store and retrieve **Node**s, which contain various-length data bytes and **Link**s to the children **Node**s. Each **Link** is a simple structure - CID of the children **Node**, name of the **Link** and total size of the serialized children **Node**.

Each **Node** is stored separately, so each children **Node** is stored in the same way, as a parent.

MerkleDAG service introduces additional type to represent graph-structure with content of all children **Node**s: **Leave**. This type is used for recursive fetch of all children **Node** with their content.

### Benefits

IPFS component for managing distributed pieces of data.

### DIFF
--- a/core/storage/ipfs/merkledag/CMakeLists.txt
+++ b/core/storage/ipfs/merkledag/CMakeLists.txt
@@ -1,0 +1,34 @@
+find_package(Protobuf REQUIRED)
+set(PB_SCHEME "merkledag.proto")
+set(PB_BUILD_DIR ${CMAKE_CURRENT_BINARY_DIR})
+
+execute_process(COMMAND ${PROTOBUF_PROTOC_EXECUTABLE} --cpp_out=${PB_BUILD_DIR} ${PB_SCHEME}
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/impl/protobuf
+    RESULT_VARIABLE CMD_OUTPUT
+    )
+
+add_library(ipfs_merkledag_service_protobuf
+    ${PB_BUILD_DIR}/merkledag.pb.h
+    ${PB_BUILD_DIR}/merkledag.pb.cc
+    )
+target_include_directories(ipfs_merkledag_service_protobuf PUBLIC ${PB_BUILD_DIR})
+target_link_libraries(ipfs_merkledag_service_protobuf
+    protobuf::libprotobuf
+    )
+disable_clang_tidy(ipfs_merkledag_service_protobuf)
+
+add_library(ipfs_merkledag_service
+    impl/link_impl.cpp
+    impl/node_impl.cpp
+    impl/merkledag_service_impl.cpp
+    impl/pb_node_encoder.cpp
+    impl/pb_node_decoder.cpp
+    impl/leaf_impl.cpp
+    )
+target_link_libraries(ipfs_merkledag_service
+    cid
+    Boost::boost
+    ipfs_blockservice
+    ipfs_datastore_in_memory
+    ipfs_merkledag_service_protobuf
+    )
--- a/core/storage/ipfs/merkledag/impl/link_impl.cpp
+++ b/core/storage/ipfs/merkledag/impl/link_impl.cpp
@@ -1,0 +1,27 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "storage/ipfs/merkledag/impl/link_impl.hpp"
+
+namespace fc::storage::ipfs::merkledag {
+  LinkImpl::LinkImpl(libp2p::multi::ContentIdentifier id,
+                     std::string name,
+                     size_t size)
+      : cid_{std::move(id)},
+        name_{std::move(name)},
+        size_{size} {}
+
+  const std::string &LinkImpl::getName() const {
+    return name_;
+  }
+
+  const CID &LinkImpl::getCID() const {
+    return cid_;
+  }
+
+  size_t LinkImpl::getSize() const {
+    return size_;
+  }
+}  // namespace fc::storage::ipfs::merkledag
--- a/core/storage/ipfs/merkledag/impl/node_impl.cpp
+++ b/core/storage/ipfs/merkledag/impl/node_impl.cpp
@@ -1,0 +1,133 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "storage/ipfs/merkledag/impl/node_impl.hpp"
+
+#include <libp2p/crypto/sha/sha256.hpp>
+#include <libp2p/multi/content_identifier_codec.hpp>
+#include <libp2p/multi/multibase_codec/codecs/base58.hpp>
+#include <libp2p/multi/multihash.hpp>
+#include "merkledag.pb.h"
+#include "storage/ipfs/merkledag/impl/pb_node_decoder.hpp"
+
+using libp2p::common::Hash256;
+using libp2p::multi::ContentIdentifierCodec;
+using libp2p::multi::HashType;
+using libp2p::multi::MulticodecType;
+using libp2p::multi::Multihash;
+using merkledag::pb::PBLink;
+using merkledag::pb::PBNode;
+using Version = libp2p::multi::ContentIdentifier::Version;
+
+namespace fc::storage::ipfs::merkledag {
+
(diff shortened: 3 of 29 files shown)


core/storage/ipfs/merkledag/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+find_package(Protobuf REQUIRED)
+set(PB_SCHEME "merkledag.proto")
```
masterjedy: Maybe it's ok for now, but pls consider moving to the way protobuf used [here](https://github.com/soramitsu/libp2p/blob/942d3cbe4387af2020ae11f8abbdae8e6a2b0448/cmake/functions.cmake#L93) 
and [here](https://github.com/soramitsu/libp2p/blob/38cf8c4f99617b6076f8ab978bd8acce7af61329/src/crypto/protobuf/CMakeLists.txt#L6)

core/storage/ipfs/merkledag/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+find_package(Protobuf REQUIRED)
```
masterjedy: this line can be moved to global config

core/storage/ipfs/merkledag/impl/leave_impl.hpp
```diff
@@ -0,0 +1,46 @@
...
+     * @brief Insert children leave
+     * @param name - id of the leave
+     * @param children - leave to insert
+     * @return operation result
+     */
+    outcome::result<void> insertSubLeave(std::string name, LeaveImpl children);
+
+   private:
+    common::Buffer content_;
+    std::map<std::string, LeaveImpl, std::less<>> children_;
```
masterjedy: why do we need to specify a standard comparator?

core/storage/ipfs/merkledag/impl/node_impl.cpp
```diff
@@ -0,0 +1,133 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "storage/ipfs/merkledag/impl/node_impl.hpp"
+
+#include <libp2p/crypto/sha/sha256.hpp>
+#include <libp2p/multi/content_identifier_codec.hpp>
+#include <libp2p/multi/multibase_codec/codecs/base58.hpp>
+#include <libp2p/multi/multihash.hpp>
+#include "merkledag.pb.h"
```
masterjedy: specify full path

core/storage/ipfs/merkledag/impl/node_impl.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+    static std::shared_ptr<Node> createFromString(const std::string &content);
+
+    static outcome::result<std::shared_ptr<Node>> createFromRawBytes(
+        gsl::span<const uint8_t> input);
+
+   private:
+    mutable boost::optional<CID> cid_{};
+    common::Buffer content_;
+    std::map<std::string, LinkImpl> links_;
```
masterjedy: why do we need to use LinkImpl instead of Link interface?

core/storage/ipfs/merkledag/impl/node_impl.cpp
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  const common::Buffer &NodeImpl::content() const {
+    return content_;
+  }
+
+  outcome::result<void> NodeImpl::addChild(const std::string &name,
+                                           std::shared_ptr<const Node> node) {
+    LinkImpl link{node->getCID(), name, node->size()};
+    std::pair<std::string, LinkImpl> data{name, std::move(link)};
```
masterjedy: try to use std::make_pair() to simplify

core/storage/ipfs/merkledag/impl/node_impl.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+    std::vector<std::reference_wrapper<const Link>> getLinks() const override;
+
+    static std::shared_ptr<Node> createFromString(const std::string &content);
+
+    static outcome::result<std::shared_ptr<Node>> createFromRawBytes(
+        gsl::span<const uint8_t> input);
+
+   private:
+    mutable boost::optional<CID> cid_{};
```
masterjedy: :'-(

core/storage/ipfs/merkledag/impl/node_impl.hpp
```diff
@@ -0,0 +1,67 @@
...
+    common::Buffer content_;
+    std::map<std::string, LinkImpl> links_;
+    PBNodeEncoder pb_node_codec_;
+    size_t child_nodes_size_{};
+    mutable boost::optional<common::Buffer> pb_cache_{boost::none};
```
masterjedy: :'-(

core/storage/ipfs/merkledag/impl/pb_node_encoder.hpp
```diff
@@ -0,0 +1,99 @@
...
+#define FILECOIN_STORAGE_IPFS_MERKLEDAG_PB_NODE_ENCODER
+
+#include <map>
+#include <memory>
+#include <string>
+
+#include <gsl/span>
+#include "common/buffer.hpp"
+#include "common/outcome.hpp"
+#include "merkledag.pb.h"
```
masterjedy: pls avoid including generated files in headers if possible, and use full path

core/storage/ipfs/merkledag/impl/link_impl.cpp
```diff
@@ -0,0 +1,27 @@
...
+namespace fc::storage::ipfs::merkledag {
+  LinkImpl::LinkImpl(libp2p::multi::ContentIdentifier id,
+                     std::string name,
+                     size_t size)
+      : cid_{std::move(id)},
+        name_{std::move(name)},
+        size_{size} {}
+
+  const std::string &LinkImpl::getName() const {
+    return this->name_;
```
Alexey-N-Chernyshov: Why do you need `this`?

core/storage/ipfs/merkledag/leave.hpp
```diff
@@ -0,0 +1,58 @@
...
+#define FILECOIN_STORAGE_IPFS_MERKLEDAG_LEAVE_HPP
+
+#include <functional>
+#include <string_view>
+
+#include "common/buffer.hpp"
+#include "common/outcome.hpp"
+
+namespace fc::storage::ipfs::merkledag {
+  class Leave {
```
Alexey-N-Chernyshov: Leave -> Leaf

core/storage/ipfs/merkledag/impl/leave_impl.hpp
sergkaprovich: For support searching in map by std::string_view
[https://stackoverflow.com/questions/35525777/use-of-string-view-for-map-lookup](https://stackoverflow.com/questions/35525777/use-of-string-view-for-map-lookup)

core/storage/ipfs/merkledag/impl/node_impl.cpp
sergkaprovich: It's impossible, because this file is generated by protobuf in the temporary build directory during cmake-configure operation.

core/storage/ipfs/merkledag/impl/node_impl.hpp
sergkaprovich: We can't instantiate abstract class _Link_

core/storage/ipfs/merkledag/impl/node_impl.cpp
sergkaprovich: Simplified to `links_.emplace(name, std::move(link));` instead `links_.emplace(std::make_pair...)`

core/storage/ipfs/merkledag/impl/node_impl.hpp
sergkaprovich: We should support general interface _fc::storage::ipfs::Block_, which instances often will be transferred by constant reference or pointer, so `Block::getCID()` can't be non-const. On the other hand - we can't recalculate CID every time was added new children link or modified content, because it's a large overhead. So, _mutable_ is very useful here.

core/storage/ipfs/merkledag/impl/link_impl.cpp
sergkaprovich: Old habit)

core/storage/ipfs/merkledag/impl/node_impl.hpp
masterjedy: You are right, sorry.

PR #78, opened 2020-01-30
Extend CID copy/move operations

sergkaprovich:
### Description of the Change

Clang-Tidy requires mark "explicit" constructors, which have one argument.
Defining `explicit CID(const ContentIdentifier &cid)` caused the need to define additional copy/move constructors and assignment operators.

### DIFF
--- a/core/primitives/cid/cid.hpp
+++ b/core/primitives/cid/cid.hpp
@@ -16,13 +16,35 @@ namespace fc {
     using ContentIdentifier::ContentIdentifier;
 
     /**
-     * ContentIdentifier is not default-constructible, but in some cases we need
+     * ContentIdentifier is not default-constructable, but in some cases we need
      * default value. This value can be used to initialize class member or local
      * variable. Trying to CBOR encode this value will yield error, to ensure
      * proper initialization.
      */
     CID();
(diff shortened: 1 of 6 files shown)


core/primitives/cid/cid.hpp
```diff
@@ -16,24 +16,46 @@ namespace fc {
     using ContentIdentifier::ContentIdentifier;
 
     /**
-     * ContentIdentifier is not default-constructible, but in some cases we need
+     * ContentIdentifier is not default-constructable, but in some cases we need
```
masterjedy: as I see in translation services, both are valid

core/primitives/cid/cid.hpp
```diff
@@ -16,24 +16,46 @@ namespace fc {
...
+    CID &operator=(const ContentIdentifier &cid);
+
+    CID &operator=(ContentIdentifier &&cid);
+
     /**
      * @brief string-encodes cid
      * @return encoded value or error
      */
-    outcome::result<std::string> toString() const;
+    [[nodiscard]] outcome::result<std::string> toString() const;
```
masterjedy: nodiscard is not really required

PR #86, opened 2020-02-05
AMT

turuslan:
### Description of the Change
AMT data structure, which is sparse array or map with int keys.
Refactor HAMT.

### Benefits
AMT for Array, MultiMap, RewardActor, MinerActor.

### Possible Drawbacks 
Behavior may be slightly different from error-prone go implementation

### DIFF
--- a/test/core/storage/amt/amt_test.cpp
+++ b/test/core/storage/amt/amt_test.cpp
@@ -20,6 +20,10 @@ using fc::storage::ipfs::InMemoryDatastore;
 
 class AmtTest : public ::testing::Test {
  public:
+  auto getRoot() {
+    return store->getCbor<Root>(amt.flush().value()).value();
+  }
+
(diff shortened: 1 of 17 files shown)


test/core/storage/amt/amt_test.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+  n.items = Node::Links{{3, "010000020000"_cid}};
+  expectEncodeAndReencode(n, "83410881d82a470001000002000080"_unhex);
+
+  n.items = Node::Links{{3, Node::Ptr{}}};
+  EXPECT_OUTCOME_ERROR(AmtError::EXPECTED_CID, encode(n));
+}
+
+TEST_F(AmtTest, SetRemoveRootLeaf) {
+  auto key = 3llu;
```
igor-egorov: it was tricky to recognize ull

PR #90, opened 2020-02-07
Feature/refactor chainepoch

Alexey-N-Chernyshov:
### Description of the Change
Change ChaneEpoch type to UBigInt
Use [boost::operators](https://www.boost.org/doc/libs/1_65_0/libs/utility/operators.htm#modable1) for BigInt.
Warning fixes.
Fix randomness_provider types.


core/clock/impl/chain_epoch_clock_impl.cpp
```diff
@@ -16,7 +16,7 @@ namespace fc::clock {
   outcome::result<ChainEpoch> ChainEpochClockImpl::epochAtTime(
       const Time &time) const {
     if (time < genesis_time_) {
-      return EpochAtTimeError::BEFORE_GENESIS;
+      return outcome::failure(EpochAtTimeError::BEFORE_GENESIS);
```
igor-egorov: why? it used to be cast to failure already.

core/crypto/randomness/impl/randomness_provider_imp.cpp
```diff
@@ -12,23 +12,24 @@ namespace fc::crypto::randomness {
...
                                                       Serialization s,
                                                       const ChainEpoch &index) {
-    return deriveRandomnessInternal(tag, std::move(s), index);
+    return deriveRandomnessInternal(
+        tag, std::move(s), index.convert_to<int64_t>());
   }
 
   Randomness RandomnessProviderImpl::deriveRandomnessInternal(
-      DomainSeparationTag tag, Serialization s, const ChainEpoch &index) {
+      DomainSeparationTag tag, Serialization s, int64_t index) {
```
igor-egorov: Why not leave ChainEpoch as it was before?
Ofc, when ChainEpoch is uint64_t, it would be better to remove const &.

core/crypto/randomness/impl/randomness_provider_imp.hpp
```diff
@@ -28,7 +28,7 @@ namespace fc::crypto::randomness {
    private:
     Randomness deriveRandomnessInternal(DomainSeparationTag tag,
                                         Serialization s,
-                                        const ChainEpoch &index);
+                                        int64_t index);
```
igor-egorov: I prefer to bring ChanEpoch type alias back for readability and maintainability purposes. (just wihtou const &)

core/primitives/chain_epoch.hpp
```diff
@@ -15,7 +14,7 @@ namespace fc::primitives {
    * @brief epoch index type represents a round of a blockchain protocol, which
    * acts as a proxy for time within the VM
    */
-  using ChainEpoch = libp2p::multi::UVarint;
+  using ChainEpoch = UBigInt;
```
igor-egorov: hmmm, but still - why we cant use ChainEpoch as argument type in functions which now takes uint64_t

core/vm/state/impl/state_tree_impl.hpp
```diff
@@ -23,7 +23,7 @@ namespace fc::vm::state {
    public:
     explicit StateTreeImpl(const std::shared_ptr<IpfsDatastore> &store);
     StateTreeImpl(const std::shared_ptr<IpfsDatastore> &store, const CID &root);
-    ~StateTreeImpl() = default;
+    ~StateTreeImpl() override = default;
```
igor-egorov: looks like when base ~ is virtual there is no need in defining default ~ in an inherited class at all

core/crypto/randomness/impl/randomness_provider_imp.cpp
Alexey-N-Chernyshov: Because this function is called with -1 as ChainEpoch.

core/crypto/randomness/impl/randomness_provider_imp.hpp
Alexey-N-Chernyshov: Either I do, but ChainEpoch is unsigned and this function is supposed to be called with chain_epoch == -1 (do not ask why).

core/clock/impl/chain_epoch_clock_impl.cpp
Alexey-N-Chernyshov: It casts to ChainEpoch now.

core/vm/state/impl/state_tree_impl.hpp
Alexey-N-Chernyshov: removed

core/primitives/chain_epoch.hpp
Alexey-N-Chernyshov: uint64_t can be cast to UBigInt, not vice versa.

core/crypto/randomness/impl/randomness_provider_imp.hpp
Alexey-N-Chernyshov: tag and index are serialized for seed, and their size should be uint64 and int64.

PR #94, opened 2020-02-12
Reward Actor and Reward Actor State

igor-egorov:
Signed-off-by: Igor Egorov igor@soramitsu.co.jp
### Description of the Change

Reward actor and its state are implemented.

### Possible Drawbacks

Error-prone arithmetic around BigInt and UBigInt.
Lacks CBOR tests.

### Usage Examples or Tests [optional]

To be done as soon as the reference golang impl gets done.


Alexey-N-Chernyshov:
There is build error on Ubuntu.
Also there are no tests, but mb it is OK since neither has specs-actor.

PR #99, opened 2020-02-17
Hamt bit width

turuslan:
### Description of the Change
Add HAMT bit width constructor parameter.
Update `keyToIndices`.

### Benefits
HAMT compatibility.

### Possible Drawbacks 
None

### DIFF
--- a/core/storage/hamt/hamt.cpp
+++ b/core/storage/hamt/hamt.cpp
@@ -33,21 +33,15 @@ namespace fc::storage::hamt {
         root_{std::make_shared<Node>()},
         bit_width_{bit_width} {}
 
-  Hamt::Hamt(std::shared_ptr<ipfs::IpfsDatastore> store)
-      : Hamt{std::move(store), kDefaultBitWidth} {}
-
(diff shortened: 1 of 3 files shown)


core/storage/hamt/hamt.hpp
```diff
@@ -107,8 +108,12 @@ namespace fc::storage::hamt {
     using Visitor = std::function<outcome::result<void>(const std::string &,
                                                         const Value &)>;
 
+    Hamt(std::shared_ptr<ipfs::IpfsDatastore> store, size_t bit_width);
```
artyom-yurin: Perhaps it is worth making a constructor with a default value

core/storage/hamt/hamt.cpp
```diff
@@ -84,6 +88,30 @@ namespace fc::storage::hamt {
     return boost::get<CID>(root_);
   }
 
+  std::vector<size_t> Hamt::keyToIndices(const std::string &key, int n) const {
```
artyom-yurin: The old function looked simpler,  did it not work correctly?

PR #105, opened 2020-02-20
Feature/graphsync

art-gor:
### Description of the Change

Graphsync protocol

### Benefits

Enables synchronizing MerkleDAG objects over the network

### Possible Drawbacks 

None


core/storage/ipfs/graphsync/impl/network/length_delimited_message_reader.cpp
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  void LengthDelimitedMessageReader::close() {
+    if (!stream_) {
+      return;
+    }
+    reading_ = false;
+    if (!stream_->isClosedForRead() && stream_.use_count() == 1) {
+      stream_->close([stream{stream_}](outcome::result<void>) {});
```
Alexey-N-Chernyshov: 1. stream is captured but not used
2. ignores possible result error in parameter

core/storage/ipfs/graphsync/impl/network/message_queue.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    if (!pending_buffers_.empty()) {
+      auto buffer = std::move(pending_buffers_.front());
+      pending_buffers_.pop_front();
+      state_.pending_bytes -= buffer->size();
+      beginWrite(std::move(buffer));
+    }
+  }
+
+  void MessageQueue::beginWrite(SharedData buffer) {
```
sergkaprovich: Pass shared pointers by reference in cases, where we don't need to share ownership, is better
https://stackoverflow.com/questions/3310737/should-we-pass-a-shared-ptr-by-reference-or-by-value

core/storage/ipfs/graphsync/impl/network/length_delimited_message_reader.cpp
art-gor: just to capture shared_ptr for stream object to be live until it closed. 
Stream logic is to be refactored in libp2p later

core/storage/ipfs/graphsync/graphsync.hpp
```diff
@@ -0,0 +1,143 @@
...
+#include <libp2p/multi/multiaddress.hpp>
+#include <libp2p/peer/peer_id.hpp>
+#include <libp2p/protocol/common/subscription.hpp>
+
+#include "common/buffer.hpp"
+#include "common/outcome.hpp"
+#include "primitives/cid/cid.hpp"
+
+namespace fc::storage::ipfs::merkledag {
+  class MerkleDagService;
```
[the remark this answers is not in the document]
art-gor: We'll revise dag<->sync feedbacks on integration to something more convenient

PR #110, opened 2020-02-23
CMake: Remove duplicate dependency

eigendude:
### Description of the Change

While compiling locally, I noticed that a dependency in `cmake/dependencies.cmake` was duplicated - it tries to find protobuf twice. This PR removes the second instance.

### Benefits

Removes unnecessary and duplicate code.

### Possible Drawbacks 

The change is in the path of the cmake project file generation, so failures are likely to be at build time instead of run time.

### Usage Examples or Tests *[optional]*

Before testing, I ran `rm -rf ~/.hunter`.

To test, I created a new build directory and ran the CMake step:

```
mkdir build
cd build
cmake ..
```

After an additional change (explained below), the cmake command completes successfully.

### Caveat

I needed a small tweak to finish "cmaking". After the hunter step finished and CMake entered the `deps` folder of git submodules, I got the following errors:

```
CMake Error at deps/filecoin-ffi.cmake:55 (target_include_directories):
  target_include_directories may only set INTERFACE properties on IMPORTED
  targets
Call Stack (most recent call first):
  deps/CMakeLists.txt:8 (include)

CMake Error at deps/filecoin-ffi.cmake:56 (target_compile_options):
  target_compile_options may only set INTERFACE properties on IMPORTED
  targets
Call Stack (most recent call first):
  deps/CMakeLists.txt:8 (include)
```

To bypass these errors, I applied the patch:

```
diff --git a/deps/filecoin-ffi.cmake b/deps/filecoin-ffi.cmake
index e7db2cc..7958d82 100644
--- a/deps/filecoin-ffi.cmake
+++ b/deps/filecoin-ffi.cmake
@@ -52,8 +52,8 @@ add_library(filecoin_ffi
         )
 
 target_link_libraries(filecoin_ffi INTERFACE ${PKG_FILECOIN_LIBRARIES})
-target_include_directories(filecoin_ffi PUBLIC ${PKG_FILECOIN_INCLUDE_DIRS})
-target_compile_options(filecoin_ffi PUBLIC ${PKG_FILECOIN_CFLAGS_OTHER})
+target_include_directories(filecoin_ffi INTERFACE ${PKG_FILECOIN_INCLUDE_DIRS})
+target_compile_options(filecoin_ffi INTERFACE ${PKG_FILECOIN_CFLAGS_OTHER})
 
 set_target_properties(filecoin_ffi PROPERTIES
         INTERFACE_INCLUDE_DIRECTORIES ${FILECOIN_FFI_PATH}/include
```

After that, `cmake ..` finished successfully. I can PR this fix, but I'm still learning why it's needed.


eigendude:
Any word?

Alexey-N-Chernyshov:
Great! I like PR that only deletes code. Thanks for your contribution!

PR #121, opened 2020-02-26
Part of updated proofs and moved sector to right place

artyom-yurin:
### Description of the Change
Moved sector to primitives
Added padded and unpadded classes
Added tools for casting Comm To CID and vice versa
Added decompose PreSealCommit

### Possible Drawbacks 
Old proofs would not work


core/primitives/cid/comm_cid.cpp
```diff
@@ -0,0 +1,118 @@
...
+    UVarint u_int_len(commitment.size());
+    auto bytes = u_int.toVector();
+    bytes.insert(bytes.end(), u_int_len.toVector().begin(), u_int_len.toVector().end());
+    bytes.insert(bytes.end(), commitment.begin(), commitment.end());
+
+    return Multihash::createFromBytes(bytes);
+  }
+
+  outcome::result<std::pair<FilecoinMultihashCode, gsl::span<const uint8_t>>>
+  readMultiHashFromBuf(gsl::span<const uint8_t> bytes) {
```
turuslan: Why not [libp2p `Multihash::createFromBytes`](https://github.com/soramitsu/libp2p/blob/master/src/multi/multihash.cpp#L65)?
Why not update libp2p impl to use uvarint size, if need it at all?

artyom-yurin: Fixed

PR #122, opened 2020-02-27
Disabled proof test

artyom-yurin:
Signed-off-by: artyom-yurin <artem_yrin@mail.ru>

### Description of the Change
Disabled proof test

### Benefits
Faster build

### DIFF
--- a/test/core/proofs/proofs_test.cpp
+++ b/test/core/proofs/proofs_test.cpp
@@ -191,7 +191,7 @@ TEST_F(ProofsTest, DISABLED_ValidPoSt) {
  * @when Commit and Verifies Seal and Unseal
  * @then success
  */
-TEST_F(ProofsTest, ValidSealAndUnseal) {
+TEST_F(ProofsTest, DISABLED_ValidSealAndUnseal) {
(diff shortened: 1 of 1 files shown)


test/core/proofs/proofs_test.cpp
```diff
@@ -191,7 +191,7 @@ TEST_F(ProofsTest, DISABLED_ValidPoSt) {
  * @when Commit and Verifies Seal and Unseal
  * @then success
  */
-TEST_F(ProofsTest, ValidSealAndUnseal) {
+TEST_F(ProofsTest, DISABLED_ValidSealAndUnseal) {
```
igor-egorov: Not clear why the test gets disabled.

Please leave a todo with Jira task id.
The comment should explain why it is disabled and when it should be enabled.

igor-egorov: Please add jira task id that describes the problem and possible solutions, which are local for CI cache or a cmake flag for enabling proofs_test target.

artyom-yurin: Added

PR #124, opened 2020-02-27
Typed actor method

turuslan:
### Description of the Change
Make actor method types strong.

### Benefits
Method params, result, number, and implementation is combined into single entity.
Runtime and runtime mock helper methods.

### Possible Drawbacks 
None


core/vm/actor/builtin/shared/shared.hpp
```diff
@@ -25,8 +24,9 @@ namespace fc::vm::actor::builtin {
    * @param miner
    * @return
    */
-  fc::outcome::result<GetControlAddressesReturn> requestMinerControlAddress(
-      Runtime &runtime, const Address &miner);
+  auto requestMinerControlAddress(Runtime &runtime, const Address &miner) {
```
Alexey-N-Chernyshov: In my opinion this change reduces readability since return type is not explicit.
> use type deduction only to make the code clearer or safer, and do not use it merely to avoid the inconvenience of writing an explicit type
https://google.github.io/styleguide/cppguide.html#Type_deduction

core/vm/actor/actor_method.hpp
```diff
@@ -9,15 +9,17 @@
...
 #include "common/outcome.hpp"
 #include "vm/actor/actor.hpp"
+#include "vm/actor/actor_encoding.hpp"
 #include "vm/exit_code/exit_code.hpp"
 #include "vm/runtime/runtime.hpp"
 
-#define ACTOR_METHOD(name)                                 \
-  outcome::result<InvocationOutput> name(Runtime &runtime, \
-                                         const MethodParams &params)
+#define ACTOR_METHOD_STUB() \
```
Alexey-N-Chernyshov: ACTOR_METHOD_DECL

Alexey-N-Chernyshov: Please, add comments.

PR #128, opened 2020-03-08
Cross compiling fixes, improvements and additions

eigendude:
## Description of the Change

This PR contains the build system fixes, improvements and additions broken out from https://github.com/filecoin-project/cpp-filecoin/pull/114.

### Structure of the PR

The PR is organized as follows:

1. Fixes needed for cross-compiling
2. Improvements needed for cross-compiling

### Fixes for cross-compiling

Commits:

- ~~CMake: Fix error including generated protobuf headers~~
- ~~Fix tinycbor include directory~~

These commits come first, because they address what could be considered defects. Inclusion would be valuable even if the rest of the PR is omitted. Further description is in the commit message.

### Changes for cross-compiling

Commits:

- [CMake: Expose CMake options to includes](https://github.com/eigendude/cpp-filecoin/commit/cross-compile-fixes~4)
- [CMake: Exclude test-related depends when building tests](https://github.com/eigendude/cpp-filecoin/commit/cross-compile-fixes~3)
- [CMake: Allow excluding git submodule depends](https://github.com/eigendude/cpp-filecoin/commit/cross-compile-fixes~2)

These changes enable more control over the build process when embedded in another build system.

CMake is controlled by user variables; the first one we need is `-DHUNTER_ENABLED=OFF` so that we can have control over dependencies.

Another useful variable is `-DTESTING=OFF`, because embedding test infra drastically increases the up-front effort. The first two commits above allow this variable to exclude test infra from the dependency discovery in addition to the build.

Finally, I introduced a new variable: `-DBUILD_INTERNAL_DEPS=OFF`. This causes the in-tree dependencies to be skipped. I needed this for a build system that doesn't support git submodules. Instead, the dependencies are provided by the build sytsem.

## Dependencies

- [x] https://github.com/soramitsu/libp2p/pull/59 needed for libp2p dependency.

## How has this been tested?

Temporarily disabled proofs and ran `make` from Kodi's build deps. Build succeeds.

Ran `make install`:

```
make[2]: *** No rule to make target 'install'.  Stop.
```

Next PR will address this.

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -1,9 +1,11 @@
 # hunter dependencies
 # https://docs.hunter.sh/en/latest/packages/
 
-# https://docs.hunter.sh/en/latest/packages/pkg/GTest.html
-hunter_add_package(GTest)
-find_package(GTest CONFIG REQUIRED)
+if (TESTING)
+  # https://docs.hunter.sh/en/latest/packages/pkg/GTest.html
+  hunter_add_package(GTest)
+  find_package(GTest CONFIG REQUIRED)
+endif()
 
 hunter_add_package(libarchive)
 find_package(libarchive CONFIG REQUIRED)
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -50,9 +50,6 @@ find_package(leveldb CONFIG REQUIRED)
 hunter_add_package(libp2p)
 find_package(libp2p CONFIG REQUIRED)
 
-hunter_add_package(c-ares)
-find_package(c-ares CONFIG REQUIRED)
-
 hunter_add_package(soralog)
 find_package(soralog CONFIG REQUIRED)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,18 +34,8 @@ endif ()
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
-include(CheckCXXCompilerFlag)
-include(cmake/toolchain-util.cmake)
-include(cmake/dependencies.cmake)
-include(cmake/functions.cmake)
-include(cmake/san.cmake)
-
 # export compile commands for clang-tidy to analyse only changed files
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-
-print("C flags: ${CMAKE_C_FLAGS}")
-print("CXX flags: ${CMAKE_CXX_FLAGS}")
-print("Using CMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}")
 
 option(TESTING "Build tests" ON)
 option(TESTING_PROOFS "Build proofs tests" OFF)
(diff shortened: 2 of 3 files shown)


cmake/dependencies.cmake
```diff
@@ -1,26 +1,32 @@
...
+  find_package(GMock CONFIG REQUIRED)
+endif()
 
 # https://docs.hunter.sh/en/latest/packages/pkg/Boost.html
 hunter_add_package(Boost COMPONENTS date_time filesystem random system)
 find_package(Boost CONFIG REQUIRED date_time filesystem random system)
 
 # https://docs.hunter.sh/en/latest/packages/pkg/Microsoft.GSL.html
 hunter_add_package(Microsoft.GSL)
-find_package(Microsoft.GSL CONFIG REQUIRED)
```
igor-egorov: According to the Hunter [docs](https://hunter.readthedocs.io/en/latest/index.html) (or [here](https://hunter.readthedocs.io/en/latest/quick-start/boost-components.html)), `find_package` is still required.

Could you please check if the code below works in your environment?
(let's keep find_package without CONFIG)
```diff
-find_package(Microsoft.GSL CONFIG REQUIRED)
+find_package(Microsoft.GSL REQUIRED)
```

eigendude: I remember trying this line with no luck. CMake has no `FindMicrosoft.GSL.cmake`, but if I add an empty one, it works. The same with cppcodec - we don't use the output `CPPCODEC_FOUND` anywhere, I just added the pkg-config because I noticed cppcodec provided pkg-config files.

Does Hunter do any magic by providing found package info to CMake? If I removed all find_package() calls, I suspect fuhon would still build, does this sound weird?

eigendude:
https://github.com/soramitsu/libp2p/pull/59 is declared as a dependency in the PR description. I updated the bullet to a check box so we can see that the dependency is fulfilled directly from this PR.

cmake/dependencies.cmake
igor-egorov: According to [the faq](https://github.com/cpp-pm/hunter/blob/master/docs/faq/why-do-we-need-hunter-add-package.rst), `hunter_add_package(foo)` will download and install `foo` package while `find_package(foo)` should only look up for files in read-only mode.

`find_package` is a CMake builtin function, which loads package settings. Thus, `find_package` looks to be required while even when hunter is not used.

eigendude: Dropped this commit

eigendude:
I updated https://github.com/soramitsu/libp2p/pull/59 and this PR. I removed all CONFIG commits. Also dropped a fix not needed after the rust BLS signature library was dropped in https://github.com/filecoin-project/cpp-filecoin/pull/129. Only 5 commits left in this PR now.

eigendude:
Is there more to do on the PR?

igor-egorov:
Hi @eigendude and sorry for the long waiting
It looks the time to merge the pr has come!
Could you please update the pr and resolve the merge conflicts?

eigendude:
Back from the dead! I rebased the PR and compile-tested. I dropped the tinycbor header change, it's no longer needed. I also added a commit removing the c-ares CMake declaration, as I can't find a use throughout the codebase.

Alexey-N-Chernyshov (approved):

PR #134, opened 2020-03-16
Api

turuslan:
### Description of the Change
Subset of jsonrpc api required for miner.
Json encoding/decoding for required types.

### Benefits
Required for go miner integration

### Possible Drawbacks 
None

### DIFF
--- a/core/crypto/bls/bls_types.hpp
+++ b/core/crypto/bls/bls_types.hpp
@@ -6,17 +6,15 @@
 #ifndef CRYPTO_BLS_PROVIDER_TYPES_HPP
 #define CRYPTO_BLS_PROVIDER_TYPES_HPP
 
-#include <filecoin-ffi/filecoin.h>
 #include <array>
 
 #include "common/outcome.hpp"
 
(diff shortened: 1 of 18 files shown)


core/crypto/bls/bls_types.hpp
```diff
@@ -6,17 +6,15 @@
...
 #include <array>
 
 #include "common/outcome.hpp"
 
 namespace fc::crypto::bls {
-
-  using PrivateKey = std::array<uint8_t, PRIVATE_KEY_BYTES>;
-  using PublicKey = std::array<uint8_t, PUBLIC_KEY_BYTES>;
-  using Signature = std::array<uint8_t, SIGNATURE_BYTES>;
-  using Digest = std::array<uint8_t, DIGEST_BYTES>;
```
artyom-yurin: If so bls_types should be removed and filecoinf-ffi dependency should be moved to bls_provider

artyom-yurin: in CMakeLists

PR #137, opened 2020-03-20
Market actor related

turuslan:
### Description of the Change
- Market actor types.
- Fix `RuntimeImpl.getImmediateCaller` returned origin.
- Piece size cbor encoding.
- `EmptyValue` with cbor encoding.
- Extract `UvarintKey` (used as `ChainEpochKey`) and `AddressKey` to "adt".
- `Map` with value and key type, and cbor encoding.
- `Array` (unused old implementation removed) with value type, and cbor encoding.
- `BalanceTable` (with behavior from specs-actors, differs from existing `BalanceTableHamt`) with cbor encoding.
- Refactor power actor.

### Benefits
Requred for market actor

### Possible Drawbacks 
None


core/adt/array.hpp
```diff
@@ -6,61 +6,86 @@
...
 
 namespace fc::adt {
+  using storage::amt::Amt;
+  using Ipld = storage::ipfs::IpfsDatastore;
 
-  /**
-   * Container for storing values preserving their order of insertion.
-   * Implementation is based on Array Mapped Trie.
-   */
+  template <typename Value>
```
Alexey-N-Chernyshov: please, restore comment

core/adt/array.hpp
```diff
@@ -6,61 +6,86 @@
...
-    template <typename T>
-    outcome::result<void> appendCbor(const T &value) {
-      OUTCOME_TRY(bytes, codec::cbor::encode(value));
-      return append(bytes);
+    using Key = uint64_t;
+    using Visitor = std::function<outcome::result<void>(Key, const Value &)>;
+
+    Array() : amt{nullptr} {}
+
+    // NOLINTNEXTLINE(google-explicit-constructor)
```
Alexey-N-Chernyshov: May be it is better to make it explicit?

core/adt/array.hpp
```diff
@@ -6,61 +6,86 @@
...
+      return amt.setCbor(key, value);
+    }
+
+    outcome::result<void> remove(Key key) {
+      return amt.remove(key);
     }
 
-    /// Iterate over stored elements acquiring their index
-    outcome::result<void> visit(const IndexedVisitor &visitor);
+    outcome::result<void> flush() {
```
Alexey-N-Chernyshov: I don't like an idea of separation of flushing and getting root_cid. I find previous behaviour is more explicit.

core/vm/actor/builtin/storage_power/storage_power_actor_state.hpp
```diff
@@ -253,6 +254,10 @@ namespace fc::vm::actor::builtin::storage_power {
 
     outcome::result<Power> getTotalNetworkPower() const;
 
+    outcome::result<Claim> assertHasClaim(const Address &address) const;
```
Alexey-N-Chernyshov: No need to export internal functions to interface.
You can make them as functions in .cpp or at least private.

test/core/adt/array_test.cpp
```diff
@@ -1,104 +0,0 @@
-/**
```
Alexey-N-Chernyshov: Why to delete all tests?

test/core/vm/actor/builtin/market/test.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "testutil/cbor.hpp"
+
+using fc::primitives::address::Address;
+using fc::primitives::piece::PaddedPieceSize;
+using fc::vm::actor::builtin::market::ClientDealProposal;
+using fc::vm::actor::builtin::market::DealProposal;
+using fc::vm::actor::builtin::market::DealState;
+using fc::vm::actor::builtin::market::State;
+
+TEST(MarketActorCborTest, State) {
```
Alexey-N-Chernyshov: please, add comments

test/core/vm/actor/builtin/market/test.cpp
```diff
@@ -0,0 +1,53 @@
...
+  expectEncodeAndReencode(
+      State{
+          .proposals = "010001020001"_cid,
+          .states = "010001020002"_cid,
+          .escrow_table = "010001020003"_cid,
+          .locked_table = "010001020004"_cid,
+          .next_deal = 1,
+          .deals_by_party = "010001020005"_cid,
+      },
+      "86d82a4700010001020001d82a4700010001020002d82a4700010001020003d82a470001000102000401d82a4700010001020005"_unhex);
```
Alexey-N-Chernyshov: Is it from go?

turuslan: Yes

core/adt/map.hpp
```diff
@@ -0,0 +1,105 @@
...
+    Hamt hamt;
+  };
+
+  template <class Stream,
+            typename V,
+            typename K,
+            size_t bit_width,
+            typename = std::enable_if_t<
+                std::remove_reference_t<Stream>::is_cbor_encoder_stream>>
+  Stream &operator<<(Stream &&s, const Map<V, K, bit_width> &map) {
```
igor-egorov: Please add a docstring here and for the operator>>.

Please use full names for Value and Keyer for type names.
Otherwise, this requires additional time while reading because (1) V comes before K (which usually stands for key and the order is opposite), and (2) K is not the key, but keyer.

core/adt/array.hpp
igor-egorov: why not to state `explicit` constructor?

core/adt/array.hpp
```diff
@@ -6,61 +6,86 @@
...
+      return amt.visit([&](auto key, auto &value) -> outcome::result<void> {
+        OUTCOME_TRY(value2, codec::cbor::decode<Value>(value));
+        return visitor(key, value2);
+      });
+    }
 
-   private:
-    std::shared_ptr<storage::ipfs::IpfsDatastore> store_;
-    storage::amt::Amt amt_;
+    storage::amt::Amt amt;
```
igor-egorov: why did you decide to expose `amt` field?

core/adt/array.hpp
igor-egorov: Why did we change the way of use of `flush`? Why this does not return cid anymore?

core/adt/address_key.hpp
```diff
@@ -0,0 +1,21 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef CPP_FILECOIN_ADT_ADDRESS_KEY_HPP
+#define CPP_FILECOIN_ADT_ADDRESS_KEY_HPP
+
+#include "primitives/address/address.hpp"
+
+namespace fc::adt {
+  struct AddressKey {
```
igor-egorov: Please check uvarint_key.hpp for comments.

core/adt/impl/multimap.cpp
```diff
@@ -5,9 +5,10 @@
 
 #include "adt/multimap.hpp"
 
-#include "adt/array.hpp"
+#include "storage/amt/amt.hpp"
 
 namespace fc::adt {
+  using storage::amt::Amt;
 
   Multimap::Multimap(const std::shared_ptr<storage::ipfs::IpfsDatastore> &store)
```
igor-egorov: Why did you changed the implementation but left `flush` untouched (which still returns cid) at the same time when Array::flush was reimplemented too?

core/adt/uvarint_key.hpp
```diff
@@ -0,0 +1,27 @@
...
+#define CPP_FILECOIN_ADT_UVARINT_KEY_HPP
+
+#include <string>
+
+#include "common/outcome.hpp"
+
+namespace fc::adt {
+  enum class UvarintKeyError { DECODE_ERROR = 1 };
+
+  struct UvarintKey {
```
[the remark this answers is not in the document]
turuslan: Then `UvarintKey`**er**.
Map is using keyer type to get key type with encoding/decoding.

turuslan:
About `cid()` and `flush()`.
You never need to call `flush()` or `cid()`.
Except when you want to commit mutations and save root cid somewhere.
Then you call `flush()` to flush cached tree to storage and put amt/hamt in "immutable" state, when it only stores root cid.
Most likely you will serialize some object that owns field of amt/hamt type, so you will need to get root cid.
However you cannot `flush()` during serialization. That's why there is separate `cid()` method.
`cid()` won't return outdated root cid. It will just throw error, which indicates that you didn't call `flush()` after mutations.

Storing plain cid in your objects, then creating amt/hamt/etc from it, then flushing to get new cid is error prone. So new `Array` and `Map` were introduced.
They are strongly typed on key and value types, and they behave like they are cid.

This PR keeps old `flush()` that returns cid, `BalanceTableHamt` and `Multimap`. That's because some code, for example power actor, depends on them and requires refactoring (too many lines for this PR).

test/core/adt/array_test.cpp
turuslan: They became irrelevant, as multimap is the only user of "append", and multimap tests cover that behavior.

core/adt/array.hpp
turuslan: Defining template cbor encoding/decoding methods as friends will fail, because they use default template parameter (`typename = enable_if_t<...>`).

Also this class is a wrapper, so amt should be accessible for those who need it.

PR #142, opened 2020-03-23
Added disable flag

artyom-yurin:
### Description of the Change
Added flag to enable proofs tests

### Benefits
You can enable tests that work for a long time, if necessary.

### Possible Drawbacks 
No disabled tests


igor-egorov:
Please use a more descriptive commit title for squash result.

PR #145, opened 2020-03-25
Market actor methods

turuslan:
### Description of the Change
Market actor methods and tests.
Fix `Amt.set` bug (values were not overwritten).
Fix `BalanceTable` bug (fail on missing key).

### Benefits
Market actor.

### Possible Drawbacks 
None

### DIFF
--- a/core/adt/balance_table.cpp
+++ b/core/adt/balance_table.cpp
@@ -14,19 +14,19 @@ OUTCOME_CPP_DEFINE_CATEGORY(fc::adt, BalanceTableError, e) {
 }
 
 namespace fc::adt {
-  outcome::result<TokenAmount> BalanceTable::get(const Key &key) {
-    OUTCOME_TRY(value, tryGet(key));
-    if (value) {
-      return *value;
-    }
-    return 0;
-  }
-
   outcome::result<void> BalanceTable::add(const Key &key, TokenAmount amount) {
     OUTCOME_TRY(value, get(key));
     value += amount;
     OUTCOME_TRY(set(key, value));
     return outcome::success();
+  }
+
+  outcome::result<void> BalanceTable::addCreate(const Key &key, TokenAmount amount) {
+    OUTCOME_TRY(value, tryGet(key));
+    if (value) {
+      amount += *value;
+    }
+    return set(key, amount);
   }
 
   outcome::result<TokenAmount> BalanceTable::subtractWithMin(const Key &key,
--- a/core/vm/actor/builtin/market/actor.cpp
+++ b/core/vm/actor/builtin/market/actor.cpp
@@ -5,7 +5,434 @@
 
 #include "vm/actor/builtin/market/actor.hpp"
 
+#include "vm/actor/builtin/market/policy.hpp"
+#include "vm/actor/builtin/shared/shared.hpp"
+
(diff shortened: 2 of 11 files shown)


core/adt/balance_table.cpp
```diff
@@ -14,21 +14,21 @@ OUTCOME_CPP_DEFINE_CATEGORY(fc::adt, BalanceTableError, e) {
...
 
 namespace fc::adt {
-  outcome::result<TokenAmount> BalanceTable::get(const Key &key) {
-    OUTCOME_TRY(value, tryGet(key));
-    if (value) {
-      return *value;
-    }
-    return 0;
-  }
-
```
artyom-yurin: Why it was removed?

core/vm/actor/builtin/market/actor.cpp
```diff
@@ -5,7 +5,434 @@
...
+    return state.locked_table.subtract(address, amount);
+  }
+
+  outcome::result<void> slashBalance(State &state,
+                                     const Address &address,
+                                     TokenAmount amount) {
+    if (amount < 0) {
+      return VMExitCode::MARKET_ACTOR_ILLEGAL_STATE;
+    }
+    OUTCOME_TRY(state.escrow_table.subtract(address, amount));
```
artyom-yurin: In the reference implementation, the amount be subtracted from the locked table only

core/vm/actor/builtin/market/actor.cpp
```diff
@@ -5,7 +5,434 @@
...
+      auto &deal = proposal.proposal;
+      if (deal.provider != provider && deal.provider != providerRaw) {
+        return VMExitCode::MARKET_ACTOR_ILLEGAL_ARGUMENT;
+      }
+
+      OUTCOME_TRY(client, runtime.resolveAddress(deal.client));
+      deal.provider = provider;
+      deal.client = client;
+
+      OUTCOME_TRY(slashed1,
```
artyom-yurin:
```suggestion
      OUTCOME_TRY(client_slashed,
```

core/vm/actor/builtin/market/actor.cpp
```diff
@@ -5,7 +5,434 @@
...
+                  updatePendingDealStatesForParty(runtime, state, client));
+      OUTCOME_TRY(slashed2,
```
artyom-yurin:
```suggestion
      OUTCOME_TRY(provider_slashed,
```

core/adt/balance_table.cpp
turuslan: In previous PR I introduced `BalanceTable` as a wrapper for `Map`, and because I thought that `BalanceTableHamt.get` behavior was wrong.

But I was mistaken while reading [balancetable.go](https://github.com/filecoin-project/specs-actors/blob/master/actors/util/adt/balancetable.go#L38).

I removed this new method, so it will fallback to `Map.get`.

core/vm/actor/builtin/market/actor.cpp
turuslan: No, both are subtracted, [market_state.go](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/market/market_state.go#L323)

PR #153, opened 2020-04-03
Api implementation

turuslan:
### Description of the Change
Implement api methods

### Benefits
Api implementation

### Possible Drawbacks 
Methods requiring chain notifications and message or block pubsub are not implemented.

### DIFF
--- a/core/api/make.cpp
+++ b/core/api/make.cpp
@@ -88,8 +88,18 @@ namespace fc::api {
               }
               return StorageParticipantBalance{*locked, *escrow - *locked};
             }},
-        // TODO(turuslan): FIL-165 implement method
-        .StateMarketDeals = {},
+        .StateMarketDeals = {[&](auto &tipset_key)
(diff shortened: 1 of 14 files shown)


core/api/make.cpp
```diff
@@ -33,55 +85,202 @@ namespace fc::api {
...
+
+          BlockMsg block;
+          block.header.miner = miner;
+          block.header.parents = parent.cids;
+          block.header.ticket = ticket;
+          block.header.height = height;
+          block.header.timestamp = timestamp;
+          block.header.epost_proof = proof;
+          block.header.parent_state_root = state_root;
+          block.header.parent_message_receipts = *context.receipts;
```
masterjedy, line 118: at least add assert that receipts field has a value

turuslan: `tipsetContext(parent, true)` sets `receipts`, so it has value

PR #160, opened 2020-04-16
Feature/secp256k1 recovery pubkey

Alexey-N-Chernyshov:
### Description of the Change
New implementation of Secp256k1 with pubkey recovery.


CMakeLists.txt
```diff
@@ -115,6 +115,7 @@ include_directories(
     # system includes
     deps/outcome
     deps/tinycbor/src
+    deps/libsecp256k1/include
```
masterjedy, line 118: in theory, modules should have exported their include paths so making it possible for dependants to find their inludes.
it should work except for some "special cases". probably it doesn't work with git submodules?

core/crypto/secp256k1/impl/secp256k1_provider_impl.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    if (!secp256k1_ecdsa_signature_parse_compact(
+            context_.get(), &sig, signature.data())) {
+      return Secp256k1Error::SIGNATURE_PARSE_ERROR;
+    }
+    if (!secp256k1_ec_pubkey_parse(
+            context_.get(), &pubkey, key.data(), key.size())) {
+      return Secp256k1Error::PUBKEY_PARSE_ERROR;
+    }
+    return secp256k1_ecdsa_verify(
```
masterjedy, line 104: does it return bool? if not, pls static cast it or convert other way

core/crypto/secp256k1/impl/secp256k1_sha256_provider_impl.hpp
```diff
@@ -0,0 +1,33 @@
...
+#define CPP_FILECOIN_CORE_CRYPTO_SECP256K1_IMPL_SECP_256_K_1_SHA_256_PROVIDER_IMPL_HPP
+
+#include "crypto/secp256k1/impl/secp256k1_provider_impl.hpp"
+
+namespace fc::crypto::secp256k1 {
+
+  /**
+   * Implementation of secp256k1 provider with sha256 digest function
+   */
+  class Secp256k1Sha256ProviderImpl : public Secp256k1ProviderImpl {
```
masterjedy, line 16: Am I right, that this provider can use any digest algorithm returning Hash256?
If yes, Blake fits this requirement.
So maybe make this provider a template class taking digest algorithm as parameter?

core/crypto/secp256k1/secp256k1_types.hpp
```diff
@@ -0,0 +1,72 @@
...
+  static const size_t kPrivateKeyLength = 32;
+  static const size_t kPublicKeyCompressedLength = 33;
+  static const size_t kPublicKeyUncompressedLength = 65;
+  static const size_t kMessageHashLength = 32;
+  static const size_t kSignatureLength = 65;
+
+  /**
+   * @brief Common types
+   */
+  using PrivateKey = std::array<uint8_t, kPrivateKeyLength>;
```
turuslan, line 23: Please use `fc::common::Blob`

core/crypto/secp256k1/impl/secp256k1_provider_impl.cpp
Alexey-N-Chernyshov: It returns 1 in case of success, or 0 otherwise. Added check.

core/crypto/secp256k1/impl/secp256k1_sha256_provider_impl.hpp
Alexey-N-Chernyshov: Yes, but we don't need bls hash.

PR #172, opened 2020-04-27
chain notify

masterjedy:
Signed-off-by: Yuriy Zarudniy <zarudniy@soramitsu.co.jp>

### Description of the Change
Refactored ChainStore, corrected logic, added comments.
Implement head change notifications.

### Possible Drawbacks 
rpc api part is not implemented yet
tests are not implemented yet


core/storage/chain/impl/chain_store_impl.hpp
```diff
@@ -21,57 +21,85 @@
...
+    }
+
+    /** chain randomness */
+    std::shared_ptr<ChainRandomnessProvider> createRandomnessProvider()
+        override;
+
+    /** @brief head change subscription */
+    connection_t subscribeHeadChanges(
+        const std::function<HeadChangeSignature> &subscriber) override {
+      return head_change_signal_.connect(subscriber);
```
turuslan, line 98: Should send CURRENT notification?

masterjedy: Yes, probably, I will add it.

masterjedy: added: if chain store has current tipset, notify subscriber about curent tipset

PR #175, opened 2020-04-29
FSM: update

sergkaprovich:
### Description of the Change
1. Switch type of the FSM from executable to library
2. Reusable io_context, which is needed for FSM
3. Test with example

### Benefits
Able to use FSM for storage and retrieval markets

### Possible Drawbacks 
Nope


core/host/context/host_context.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * @brief Get IO context
+     * @return current IO context
+     */
+    virtual std::shared_ptr<IoContext> getIoContext() const = 0;
+
+    /**
+     * @brief Run IO context for specified time
+     * @param seconds - time to run
```
masterjedy, line 35: specify that zero means forever

PR #182, opened 2020-04-30
State API

artyom-yurin:
### Description of the Change
API for state

### Benefits
The state now has API

### Possible Drawbacks 
None


core/api/make.cpp
```diff
@@ -310,17 +320,98 @@ namespace fc::api {
...
+            }
+
+            return true;
+          };
+
+          std::vector<CID> result;
+
+          while (static_cast<int64_t>(context.tipset.height) >= to_height) {
+            // TODO(turuslan): FIL-146 randomness from tipset
+            std::shared_ptr<RandomnessProvider> randomness;
```
turuslan, line 345: You dont need `Env` here. Just keep set of visited message cids.

core/api/make.cpp
```diff
@@ -310,17 +320,98 @@ namespace fc::api {
...
           OUTCOME_TRY(context, tipsetContext(tipset_key));
           OUTCOME_TRY(power_state, context.powerState());
           return power_state.claims.keys();
         }},
+        .StateListActors = {[=](auto &tipset_key)
+                                -> outcome::result<std::vector<Address>> {
+          OUTCOME_TRY(context, tipsetContext(tipset_key));
+          OUTCOME_TRY(root, context.state_tree.flush());
+          adt::Map<Actor, adt::AddressKeyer> actors(root);
+          actors.load(context.state_tree.getStore());
```
turuslan, line 411: Use `ipld` directly (`makeImpl` param)

core/api/make.cpp
artyom-yurin: Removed and added `std::set` for collect unique CIDs

core/api/make.cpp
artyom-yurin: Fixed

PR #187, opened 2020-05-04
market deal discovery

Alexey-N-Chernyshov:

### Description of the Change
Markets deal discovery.
Stores list of peer_info(address, peer_id) for each proposal deal cid.


core/markets/retrieval/common_types.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include "primitives/address/address.hpp"
+
+namespace fc::markets::retrieval {
+  using libp2p::peer::PeerId;
+  using primitives::address::Address;
+
+  /**
+   * @struct Everything needed to make a deal with a miner
+   */
+  struct RetrievalPeer {
```
sergkaprovich, line 19: We can use existing libp2p::peer::PeerInfo instead - it supports several addresses for each peer, maybe, it has sense (e.g. host supports tcp and udp and it will serve `/ip4/127.0.0.1/tcp/8080` and `/ip4/127.0.0.1/udp/8000`

PR #189, opened 2020-05-07
Feature/piece storage

sergkaprovich:
### Description of the Change

Piece storage implementation

### Benefits

Ability to store information about pieces location in the sectors and payload block positions in the pieces

### Possible Drawbacks 
None


core/storage/piece/impl/piece_storage_impl.cpp
```diff
@@ -0,0 +1,78 @@
...
+    OUTCOME_TRY(key, payload_cid.toString());
+    Buffer storage_key =
+        PieceStorageImpl::convertKey(kLocationPrefix, std::move(key));
+    OUTCOME_TRY(value, storage_->get(storage_key));
+    OUTCOME_TRY(payload_info, codec::cbor::decode<PayloadBlockInfo>(value));
+    return std::move(payload_info);
+  }
+
+  PieceStorageImpl::Buffer PieceStorageImpl::convertKey(std::string prefix,
+                                                        std::string key) {
```
Alexey-N-Chernyshov, line 70: Changing std::string to CID will save 3 lines 😄

sergkaprovich: Yes, I thought about it, but golang implementation uses exactly string CID representation, and we can't avoid namespaces prefixes(

PR #213, opened 2020-06-09
Add datatransfer to storage markets

Alexey-N-Chernyshov:
### Description of the Change
Update datatransfer, add datatransfer storage market validators, add datatransfer to storage markets.


core/data_transfer/impl/libp2p_data_transfer_network.cpp
```diff
@@ -4,28 +4,33 @@
...
 /**
  * Check outcome and calls receiver->receiveError in case of failure
  */
-#define CHECK_OUTCOME_RESULT(expression, receiver) \
-  if (expression.has_error()) {                    \
-    receiver->receiveError();                      \
-    stream->reset();                               \
-    return;                                        \
+#define CHECK_OUTCOME_RESULT(expression, receiver)                           \
+  auto UNIQUE_NAME(_r) = expression;                                         \
```
turuslan, line 14: Currently UNIQUE_NAME uses __LINE__, so reusing it here works now.
But it probably will change and use __COUNTER__, so this code will break.
Please pass UNIQUE_NAME as parameter to another macro.

Alexey-N-Chernyshov: done

PR #221, opened 2020-06-17
Drand manager

igor-egorov:
Implement drand beacons manager

Add debugging tools inside `drand/example` directory

https://github.com/filecoin-project/lotus/blob/v0.3.0/chain/beacon/drand/drand.go

### DIFF
--- a/core/drand/impl/beaconizer.cpp
+++ b/core/drand/impl/beaconizer.cpp
@@ -70,7 +70,7 @@ namespace fc::drand {
                                               std::move(net_key),
                                               max_cache_size);
     OUTCOME_TRY(instance->init());
-    return instance;
+    return outcome::success(std::move(instance));
   }
 
   BeaconizerImpl::BeaconizerImpl(uint64_t filecoin_genesis_time,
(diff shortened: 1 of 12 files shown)


core/drand/impl/beaconizer.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+  outcome::result<void> BeaconizerImpl::verifyNetworkKey(
+      const Bytes &key) const {
+    auto &key1 = network_key_;
+    if (key.size() != key1.size()) {
+      return crypto::bls::Errors::InvalidPublicKey;
+    }
+    if (not std::equal(key1.begin(), key1.end(), key.begin())) {
+      return Error::NETWORK_KEY_MISMATCH;
+    }
```
artyom-yurin, line 163: Maybe do via `==`

igor-egorov: not for this case because here we have std::array on the one side and std::vector on the other

core/drand/impl/beaconizer.cpp
```diff
@@ -0,0 +1,212 @@
...
+                           std::move(drand_servers),
+                           std::move(network_public_key),
+                           max_cache_size} {};
+    };
+    auto instance =
+        std::make_unique<make_unique_enabler>(filecoin_genesis_time,
+                                              filecoin_round_time,
+                                              std::move(drand_servers),
+                                              std::move(net_key),
+                                              max_cache_size);
```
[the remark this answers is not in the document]
igor-egorov: This way of unique_ptr instantiation is [discouraged](https://stackoverflow.com/questions/37514509/advantages-of-using-stdmake-unique-over-new-operator) nowadays.

PR #223, opened 2020-06-17
Use gcc 9 on github ci

igor-egorov:
bump gcc version


igor-egorov:
The default ubuntu 18.04 compiler is GCC 7.4 which gets updated via apt to 7.5. That is why it looks like we cannot bump the version.

PR #225, opened 2020-06-19
MinerGetBaseInfo

turuslan:
### Description of the Change
MinerGetBaseInfo, StateMinerSectors, StateProvingSet

### DIFF
--- a/core/api/make.cpp
+++ b/core/api/make.cpp
@@ -9,6 +9,7 @@
 #include <libp2p/peer/peer_id.hpp>
 
 #include "blockchain/production/block_producer.hpp"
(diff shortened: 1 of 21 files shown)


core/api/make.cpp
```diff
@@ -45,6 +48,21 @@ namespace fc::api {
   using vm::runtime::Env;
   using connection_t = boost::signals2::connection;
 
+  constexpr EpochDuration kWinningPoStSectorSetLookback{10};
+
+  // TODO: move
```
igor-egorov, line 53: please put better description - what to move and where

PR #227, opened 2020-06-21
Index lock

turuslan:
### Description of the Change
Index lock: https://github.com/filecoin-project/sector-storage/blob/master/stores/index_locks.go


artyom-yurin:
One remark. Maybe should to move lock_index into different cmake target

PR #237, opened 2020-07-06
Local Worker

artyom-yurin:
### Description of the Change
- Worker(interface)
- Local Worker
- Seal Tasks
- Updated Local Store

### Benefits
Scheduler is unlocked

### Possible Drawbacks 
None


core/primitives/worker_config.hpp
```diff
@@ -0,0 +1,20 @@
+/**
```
Alexey-N-Chernyshov, line 1: Imo config of an implementation is not a primitive and should be placed beside implementation (LocalWorker).

core/sector_storage/spec_interfaces/sealer.hpp
```diff
@@ -3,46 +3,36 @@
...
-
-    // SEALER
     virtual outcome::result<PreCommit1Output> sealPreCommit1(
         const SectorId &sector,
         const SealRandomness &ticket,
         gsl::span<const PieceInfo> pieces) = 0;
 
     virtual outcome::result<SectorCids> sealPreCommit2(
-        const SectorId &sector, const PreCommit1Output &pc10) = 0;
+        const SectorId &sector, const PreCommit1Output &pc1o) = 0;
```
Alexey-N-Chernyshov, line 35: It is better to use full naming instead of abbreviations.

core/sector_storage/impl/local_worker.cpp
```diff
@@ -0,0 +1,493 @@
...
+    mem_info.virtual_used = mem_info.virtual_total - mem_info.virtual_free;
+
+    if (!has_available) {
+      mem_info.available = mem_info.free + mem_info.metrics["Buffers"]
+                           + mem_info.metrics["Cached"];
+    }
+
+    result.resources.reserved_memory =
+        mem_info.virtual_used + mem_info.total - mem_info.available;
+#endif
```
Alexey-N-Chernyshov, line 398: Otherwise report platform is not supported.

PR #242, opened 2020-07-14
fixes and refactor

turuslan:
### Description of the Change
- Single randomness function.
- Fix amt (wrong fast pow, missing condition).
- "bigdiv" compatible with "big.Div" from go (rounding for negative fractions), instead of boost::multiprecision::cpp_int "operator/".
- Pricelist, charging ipld.
- Fixes from debugging.
- Remove "ActorSubstateCID".


core/primitives/big_int.hpp
```diff
@@ -12,8 +12,18 @@
 
 namespace fc::primitives {
   using BigInt = boost::multiprecision::cpp_int;
+  inline BigInt bigdiv(const BigInt &n, const BigInt &d) {
```
Alexey-N-Chernyshov, line 15: At least reasoning for it should be explained.

test/core/primitives/big_int_test.cpp
```diff
@@ -42,6 +42,9 @@ TEST(BigInt, Divide) {
   BigInt a{8};
   BigInt b{2};
   ASSERT_EQ(a / b, BigInt{4});
+  ASSERT_EQ(BigInt{-1} / BigInt{3}, BigInt{0});
+  ASSERT_EQ(fc::bigdiv(0, 3), BigInt{0});
```
Alexey-N-Chernyshov, line 46: I think it is better to create another test for bigdiv.

PR #248, opened 2020-07-23
Manager

artyom-yurin:
### Description of the Change
- Manager
- Reformat constructors and structure

### Benefits
SectorStorage is done

### Possible Drawbacks 
None


core/sector_storage/stores/impl/index_lock.cpp
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#include "sector_storage/stores/index_lock.hpp"
```
turuslan, line 6: Maybe this doesn't conform to styleguide.

PR #249, opened 2020-07-24
WindowPoStScheduler

turuslan:
### Description of the Change
https://github.com/filecoin-project/lotus/blob/master/storage/wdpost_sched.go
https://github.com/filecoin-project/lotus/blob/master/storage/wdpost_run.go

### Todo
`generateWindowPoSt`.


core/miner/windowpost.hpp
```diff
@@ -0,0 +1,32 @@
...
+  using api::Api;
+  using api::DeadlineInfo;
+  using api::Tipset;
+
+  // synchronous WindowPoSt
+  struct WindowPoStScheduler
+      : public std::enable_shared_from_this<WindowPoStScheduler> {
+    static constexpr auto kStartConfidence{4};
+
+    outcome::result<std::shared_ptr<WindowPoStScheduler>> create(
```
artyom-yurin, line 22: maybe `unique_ptr`

turuslan: No, need to capture itself in lambda.

core/miner/windowpost.cpp
```diff
@@ -0,0 +1,96 @@
...
+        OUTCOME_TRY(proof,
+                    prover->generateWindowPoSt(miner.getId(), sectors2, rand));
+        params.proofs = std::move(proof.proof);
+        OUTCOME_TRY(api->MpoolPushMessage({
+            miner,
+            worker,
+            0,
+            1000,
+            1,
+            10000000,
```
Alexey-N-Chernyshov, line 87: Please, use constants for gas values. We have kDefaultGasPrice and kDefaultGasLimit, but they are different.

PR #254, opened 2020-07-30
Mining

turuslan:
### Description of the Change
Block mining cycle
https://github.com/filecoin-project/lotus/blob/master/miner/miner.go

### DIFF
--- a/core/api/api.hpp
+++ b/core/api/api.hpp
@@ -33,6 +33,7 @@
 
 #define API_METHOD(_name, _result, ...)                                    \
   struct _##_name : std::function<outcome::result<_result>(__VA_ARGS__)> { \
+    using function::function;                                              \
     using Result = _result;                                                \
     using Params = ParamsTuple<__VA_ARGS__>;                               \
     static constexpr auto name = "Filecoin." #_name;                       \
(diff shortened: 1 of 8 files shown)


core/api/api.hpp
```diff
@@ -196,6 +197,10 @@ namespace fc::api {
     SectorSize sector_size;
     BeaconEntry prev_beacon;
     std::vector<BeaconEntry> beacons;
+
+    auto &beacon() const {
+      return beacons.empty() ? prev_beacon : *beacons.rbegin();
```
Alexey-N-Chernyshov, line 202:
```suggestion
      return beacons.empty() ? prev_beacon : beacons.back();
```

core/miner/mining.hpp
```diff
@@ -0,0 +1,79 @@
...
+    }
+    std::shared_ptr<io_context> io;
+    std::unique_ptr<system_timer> timer;
+    std::shared_ptr<Api> api;
+    std::shared_ptr<Prover> prover;
+    Address miner;
+    boost::optional<Tipset> ts;
+    BigInt weight;
+    size_t skip{};
+    std::unordered_set<std::pair<TipsetKey, size_t>, pair_hash> mined;
```
Alexey-N-Chernyshov, line 63: Not sure that all members should be public.

PR #274, opened 2020-09-18
updates

turuslan:
### Description of the Change
- update blocksync
- update market and power actor states
- add const file
- update weight and fee function
- update some api
- add pubsub and empty main


core/markets/storage/client/impl/storage_market_client_impl.cpp
```diff
@@ -109,7 +109,7 @@ namespace fc::markets::storage::client {
     std::vector<StorageProviderInfo> storage_providers;
     for (const auto &miner_address : miners) {
       OUTCOME_TRY(miner_info, api_->StateMinerInfo(miner_address, tipset_key));
-      PeerInfo peer_info{.id = miner_info.peer_id, .addresses = {}};
+      PeerInfo peer_info{.id = *miner_info.peer_id, .addresses = {}};
```
artyom-yurin, line 112: What if miner_info is none?

artyom-yurin (approved):

core/markets/storage/client/impl/storage_market_client_impl.cpp
turuslan: Don't know, lotus has no check there.

core/api/api.hpp
```diff
@@ -270,6 +272,16 @@ namespace fc::api {
     uint64_t partition;
   };
 
+  struct MinerInfo2 {
```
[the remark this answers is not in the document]
Alexey-N-Chernyshov: +1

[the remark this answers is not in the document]
Alexey-N-Chernyshov: it is identical to fc::vm::actor::builtin::miner::MinerInfo

core/const.cpp
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "const.hpp"
+
+#define DEFINE(x) decltype(x) x
+
+namespace fc {
+  DEFINE(kConsensusMinerMinPower){BigInt{1} << 40};
```
Alexey-N-Chernyshov, line 11: need macro for this?

turuslan: With this macro you don't need to repeat type.

Alexey-N-Chernyshov (approved):

PR #288, opened 2020-10-26
Feature/cgo invoker

Alexey-N-Chernyshov:
### Description of the Change
Create CgoInvoker with Invoker Interface.

### Benefits
Unified interface, can be replaced with builtin actors.

### Possible Drawbacks 
no


core/vm/runtime/impl/env.cpp
```diff
@@ -200,6 +201,7 @@ namespace fc::vm::runtime {
     execution->gas_limit = message.gas_limit;
     execution->origin = message.from;
     execution->origin_nonce = message.nonce;
+    execution->invoker = std::move(invoker);
```
turuslan, line 204: Don't `move` here

core/vm/actor/cgo/cgo_invoker.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include "vm/actor/invoker.hpp"
+
+namespace fc::vm::actor::cgo {
+
+  /**
+   * Invoker implementation that invokes GoLang specs-actors using cgo
+   */
+  class CgoInvoker : public Invoker {
+   public:
+    explicit CgoInvoker(bool test_vectors);
```
turuslan, line 18: Should `InvokerImpl` have same option?

core/vm/runtime/impl/env.cpp
Alexey-N-Chernyshov: ok

core/vm/actor/cgo/cgo_invoker.hpp
Alexey-N-Chernyshov: test_vectors is used only for randomness, and randomness is provided by `Runtime`. I think it should be removed from `CgoInvoker` to `Runtime`, and `InvokerImpl` shouldn't have the option.

turuslan (approved):

core/vm/actor/cgo/cgo_invoker.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+    void config(const StoragePower &min_verified_deal_size,
+                const StoragePower &consensus_miner_min_power,
+                const std::vector<RegisteredProof> &supported_proofs) override;
+
+    outcome::result<InvocationOutput> invoke(const Actor &actor,
+                                             Runtime &runtime) override;
+
+   private:
+    bool _test_vectors{false};
```
wer1st, line 28: For what? You already defined it in the constructor's initialisation list.
Moreover it is not a good practice to define members value in .h files

Alexey-N-Chernyshov: fixed

Alexey-N-Chernyshov:
reopen PR as trunk/cpp-actors

PR #307, opened 2020-12-07
Feature/cron actor v2

Alexey-N-Chernyshov:
### Description of the Change
Add cron v2 actor.


wer1st (changes requested):
Please add also cron actor unit test v2

wer1st (approved):

PR #310, opened 2020-12-10
Voucher's signature

wer1st:
### Description of the Change
Use buffer instead of Signature in Signed Voucher.
Enable 2 test vectors that were blocked by this issue.

### Benefits
Delayed Signature parsing that allows to charge gas before VerifySignature() and have the same behaviour as Lotus

### DIFF
--- a/core/vm/runtime/impl/runtime_impl.cpp
+++ b/core/vm/runtime/impl/runtime_impl.cpp
@@ -213,7 +213,7 @@ namespace fc::vm::runtime {
       const Address &address,
       gsl::span<const uint8_t> data) {
     const auto bls = Signature::isBls(signature_bytes);
-    if (!bls) {
+    if (bls.has_error()) {
       return false;
     }
     OUTCOME_TRY(
--- a/core/vm/runtime/impl/runtime_impl.cpp
+++ b/core/vm/runtime/impl/runtime_impl.cpp
@@ -51,7 +51,8 @@ namespace fc::vm::runtime {
       DomainSeparationTag tag,
       ChainEpoch epoch,
       gsl::span<const uint8_t> seed) const {
-    return execution_->env->randomness->getRandomnessFromBeacon(tag, epoch, seed);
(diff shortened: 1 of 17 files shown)


turuslan (approved):

core/vm/runtime/impl/runtime_impl.cpp
```diff
@@ -207,6 +208,32 @@ namespace fc::vm::runtime {
         .verify(account, data, signature);
   }
 
+  outcome::result<bool> RuntimeImpl::verifySignatureBytes(
+      const Buffer &signature_bytes,
+      const Address &address,
+      gsl::span<const uint8_t> data) {
+    const auto bls = Signature::isBls(signature_bytes);
+    if (!bls) {
+      return false;
```
Alexey-N-Chernyshov, line 217: Why `secp256k1` not supported?

core/vm/runtime/impl/runtime_impl.cpp
```diff
@@ -207,6 +208,32 @@ namespace fc::vm::runtime {
         .verify(account, data, signature);
   }
 
+  outcome::result<bool> RuntimeImpl::verifySignatureBytes(
+      const Buffer &signature_bytes,
+      const Address &address,
+      gsl::span<const uint8_t> data) {
+    const auto bls = Signature::isBls(signature_bytes);
+    if (!bls) {
```
Alexey-N-Chernyshov, line 216:
```suggestion
    if (bls.has_error()) {
```

core/vm/runtime/impl/runtime_impl.cpp
Alexey-N-Chernyshov: It is outcome::result check.

Alexey-N-Chernyshov (approved):

core/vm/runtime/impl/runtime_impl.cpp
wer1st: Yes

PR #329, opened 2021-01-27
State tree cache

turuslan:
Makes vm faster

https://github.com/filecoin-project/lotus/blob/master/chain/state/statetree.go#L38

### DIFF
--- a/core/vm/state/impl/state_tree_impl.cpp
+++ b/core/vm/state/impl/state_tree_impl.cpp
@@ -38,7 +38,7 @@ namespace fc::vm::state {
       if (it->removed.count(address_id.getId())) {
         return storage::hamt::HamtError::kNotFound;
       }
-      auto actor{it->actors.find(address_id.getId())};
+      const auto actor{it->actors.find(address_id.getId())};
       if (actor != it->actors.end()) {
         return actor->second;
       }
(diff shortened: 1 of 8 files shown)


core/vm/state/impl/state_tree_impl.cpp
```diff
@@ -12,32 +12,55 @@ namespace fc::vm::state {
...
   }
 
   outcome::result<Actor> StateTreeImpl::get(const Address &address) {
     OUTCOME_TRY(address_id, lookupId(address));
-    return by_id.get(address_id);
+    for (auto it{tx_.rbegin()}; it != tx_.rend(); ++it) {
+      if (it->removed.count(address_id.getId())) {
+        return storage::hamt::HamtError::kNotFound;
+      }
+      auto actor{it->actors.find(address_id.getId())};
```
wer1st, line 41:
```suggestion
      const auto actor{it->actors.find(address_id.getId())};
```

core/vm/state/impl/state_tree_impl.cpp
```diff
@@ -12,32 +12,55 @@ namespace fc::vm::state {
...
+    OUTCOME_TRY(set(address_id, actor));
+    return std::move(actor);
   }
 
   outcome::result<Address> StateTreeImpl::lookupId(const Address &address) {
     if (address.isId()) {
       return address;
     }
+    for (auto it{tx_.rbegin()}; it != tx_.rend(); ++it) {
+      auto id{it->lookup.find(address)};
```
wer1st, line 56:
```suggestion
      const auto id{it->lookup.find(address)};
```

wer1st (approved):

Alexey-N-Chernyshov (approved):

PR #330, opened 2021-01-31
vm buffered ipld

turuslan:
- enable disabled test-vectors
- fix power v2 claim type
- fix verify post
- fix duplicate error cpp file
- handle exit code
- fix state tree recursion when using dvm
- simplified batch verify seal
- cid utils
- buffered ipld, writes only final changes and not unreferenced temporary changes, also assumes cbor blake cids

### DIFF
--- a/core/primitives/cid/cid.hpp
+++ b/core/primitives/cid/cid.hpp
@@ -9,9 +9,13 @@
 #include <spdlog/fmt/fmt.h>
 #include <libp2p/multi/content_identifier.hpp>
 
+#include "common/blob.hpp"
+#include "common/buffer.hpp"
 #include "common/outcome.hpp"
 
 namespace fc {
+  using common::Hash256;
+
   class CID : public libp2p::multi::ContentIdentifier {
    public:
     using ContentIdentifier::ContentIdentifier;
@@ -75,6 +79,10 @@ namespace fc {
   };
 
   size_t hash_value(const CID &cid);
+
+  bool isCbor(const CID &cid);
+  boost::optional<Hash256> asBlake(const CID &cid);
+  boost::optional<BytesIn> asIdentity(const CID &cid);
 }  // namespace fc
 
 namespace std {
--- a/core/vm/exit_code/impl/exit_code.cpp
+++ b/core/vm/exit_code/impl/exit_code.cpp
@@ -38,7 +38,7 @@ namespace fc::vm {
 
   outcome::result<VMExitCode> asExitCode(const std::error_code &error) {
     if (isVMExitCode(error)) {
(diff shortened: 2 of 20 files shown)


core/primitives/cid/cid.hpp
```diff
@@ -75,6 +79,10 @@ namespace fc {
   };
 
   size_t hash_value(const CID &cid);
+
+  bool isCbor(const CID &cid);
+  boost::optional<Hash256> asBlake(const CID &cid);
+  boost::optional<BytesIn> asIdentity(const CID &cid);
```
wer1st, line 85:
```suggestion
  bool isCbor(const CID &cid) const;
  boost::optional<Hash256> asBlake(const CID &cid) const;
  boost::optional<BytesIn> asIdentity(const CID &cid) const;
```

core/vm/exit_code/impl/exit_code.cpp
```diff
@@ -35,4 +35,11 @@ namespace fc::vm {
   bool isAbortExitCode(const std::error_code &error) {
     return error.category() == __libp2p::Category<VMAbortExitCode>::get();
   }
+
+  outcome::result<VMExitCode> asExitCode(const std::error_code &error) {
+    if (isVMExitCode(error)) {
+      return outcome::success(VMExitCode{error.value()});
```
wer1st, line 41:
```suggestion
      return outcome::success(static_cast<VMExitCode>(error.value()));
```

core/vm/runtime/impl/runtime_impl.cpp
```diff
@@ -181,11 +181,37 @@ namespace fc::vm::runtime {
...
+    BatchSealsOut res;
+    if (batch) {
+      OUTCOME_TRY(batch->visit([&](auto &miner,
+                                   auto &seals) -> outcome::result<void> {
+        OUTCOME_TRY(count, seals.size());
+        std::vector<SectorNumber> successful;
+        successful.reserve(count);
+        std::set<SectorNumber> seen;
+        OUTCOME_TRY(seals.visit([&](auto, auto &seal) -> outcome::result<void> {
+          auto verified{proofs::Proofs::verifySeal(seal)};
```
wer1st, line 201:
```suggestion
          const auto verified{proofs::Proofs::verifySeal(seal)};
```

core/primitives/cid/cid.hpp
turuslan: They are not member functions

wer1st: Yes, I missed that

wer1st (approved):

Alexey-N-Chernyshov (approved):

PR #332, opened 2021-02-02
Return codes

wer1st:
### Description of the Change

Make actor's return codes behaviour consistent with Lotus and spec actors
Redesign Market actor v0
Redesign Payment channel actor v0 & v2
Implement Init actor v2 & v3
Implement Cron actor v3


core/proofs/impl/proofs.cpp
```diff
@@ -144,7 +144,7 @@ namespace fc::proofs {
   }
 
   // ******************
-  // TO С CASTED FUNCTIONS
+  // TO C CASTED FUNCTIONS
```
wer1st, line 147: Change Russian "С" to Eng "C"

wer1st:
At first, please take a look at exit_code.hpp and runtime.hpp

core/vm/actor/actor_encoding.hpp
```diff
@@ -34,7 +34,7 @@ namespace fc::vm::actor {
     auto maybe_params = codec::cbor::decode<T>(params_bytes);
     if (!maybe_params) {
       return outcome::failure(v7 ? VMExitCode::kErrSerialization
-                                 : VMExitCode::kErrSerializationPre7);
+                                 : VMExitCode::kOldErrActorFailure);
```
turuslan, line 37: Older name were more specific (ErrSerialization + "before v7")

wer1st: But it is more common error.
I think that it is clear here that kOldErrActorFailure error is used if v7 == false.
I could add a comment if you wish

test/core/vm/actor/builtin/v2/power/storage_power_actor_v2_test.cpp
```diff
@@ -192,7 +192,7 @@ namespace fc::vm::actor::builtin::v2::storage_power {
     constructed();
     callerCodeIdIs(kEmptyObjectCid);
 
-    EXPECT_OUTCOME_ERROR(VMExitCode::kSysErrForbidden,
+    EXPECT_OUTCOME_ERROR(ABORT_CAST(VMExitCode::kSysErrForbidden),
```
turuslan, line 195: Maybe add "EXPECT_ABORT(exit, expr)" macro for actors tests?

wer1st: Maybe later if it needed.
@Alexey-N-Chernyshov, as I remember you want to remake actor tests, you could take this idea

core/vm/exit_code/exit_code.hpp
```diff
@@ -111,6 +143,170 @@ OUTCOME_HPP_DECLARE_ERROR(fc::vm, VMFatal);
...
+                                          const VMExitCode &default_error) {
+    if (res.has_error()) {
+      if (isFatal(res.error()) || isAbortExitCode(res.error())) {
+        return res.error();
+      }
+      if (isVMExitCode(res.error())) {
+        ABORT(static_cast<VMExitCode>(res.error().value()));
+      }
+      ABORT(default_error);
+    }
```
turuslan, line 192:
```suggestion
    OUTCOME_TRY(requireNoError(res, default_error));
```

core/vm/exit_code/exit_code.hpp
```diff
@@ -111,6 +143,170 @@ OUTCOME_HPP_DECLARE_ERROR(fc::vm, VMFatal);
...
+   * Aborts execution if res is not success
+   * @tparam T - result type
+   * @param res - result to check
+   * @return If res has no error, success() returned. Otherwise VMAbortExitCode
+   * returned
+   */
+  template <typename T>
+  outcome::result<void> requireSuccess(const outcome::result<T> &res) {
+    if (res.has_error()) {
+      ABORT(res.error().value());
```
turuslan, line 206: Must check `asExitCode` before `ABORT` cast.
Any other error is returned as is.

core/vm/exit_code/exit_code.hpp
```diff
@@ -111,6 +143,170 @@ OUTCOME_HPP_DECLARE_ERROR(fc::vm, VMFatal);
...
+   * @tparam T - result type
+   * @param res - result to check
+   * @return If res has no error, res returned. Otherwise VMAbortExitCode
+   * returned
+   */
+  template <typename T>
+  outcome::result<T> requireSuccessAssign(outcome::result<T> &&res) {
+    if (res.has_error()) {
+      ABORT(res.error().value());
+    }
```
turuslan, line 222:
```suggestion
    OUTCOME_TRY(requireSuccess(res));
```

core/vm/exit_code/exit_code.hpp
```diff
@@ -111,6 +143,170 @@ OUTCOME_HPP_DECLARE_ERROR(fc::vm, VMFatal);
...
+   */
+  template <typename T>
+  outcome::result<T> changeErrorAssign(outcome::result<T> &&res,
+                                       const VMExitCode &error) {
+    if (res.has_error()) {
+      if (isFatal(res.error()) || isAbortExitCode(res.error())) {
+        return res.error();
+      }
+      return outcome::failure(error);
+    }
```
turuslan, line 264:
```suggestion
    OUTCOME_TRY(changeError(res, error));
```

core/vm/exit_code/exit_code.hpp
```diff
@@ -111,6 +143,170 @@ OUTCOME_HPP_DECLARE_ERROR(fc::vm, VMFatal);
...
+  template <typename T>
+  outcome::result<void> requireNoError(const outcome::result<T> &res,
+                                       const VMExitCode &default_error) {
+    if (res.has_error()) {
+      if (isFatal(res.error()) || isAbortExitCode(res.error())) {
+        return res.error();
+      }
+      if (isVMExitCode(res.error())) {
+        ABORT(static_cast<VMExitCode>(res.error().value()));
+      }
```
turuslan, line 165:
```suggestion
      if (auto exit{asExitCode(res)}) {
        ABORT(exit.value());
      }
```

core/vm/exit_code/exit_code.hpp
```diff
@@ -111,6 +143,170 @@ OUTCOME_HPP_DECLARE_ERROR(fc::vm, VMFatal);
...
+   */
+  template <typename T>
+  outcome::result<T> changeErrorAbortAssign(outcome::result<T> &&res,
+                                            const VMExitCode &error) {
+    if (res.has_error()) {
+      if (isFatal(res.error()) || isAbortExitCode(res.error())) {
+        return res.error();
+      }
+      ABORT(error);
+    }
```
turuslan, line 306:
```suggestion
    OUTCOME_TRY(changeErrorAbort(res, error));
```

core/vm/exit_code/exit_code.hpp
```diff
@@ -3,17 +3,70 @@
...
+/**
+ * In addition to CHANGE_ERROR_ABORT allows to create res variable
+ */
+#define CHANGE_ERROR_ABORT_A(res, expr, err_code) \
+  OUTCOME_TRY((res), changeErrorAbortAssign((expr), (err_code)))
+
+/**
+ * Return VMExitCode as VMAbortExitCode for special handling
+ */
+#define ABORT_CAST(err_code) static_cast<VMAbortExitCode>(err_code)
```
turuslan, line 64: Add method (maybe like `asExitCode`) which will only allow casting from `VMAbortExitCode` and `VMExitCode`

wer1st: Where do we need it to use?

turuslan: Where it's used.
This macros `static_cast`'s anything to `VMAbortExitCode`, this isn't right

core/vm/exit_code/exit_code.hpp
wer1st: But spec actors aborts it anyway.
`if !e.IsSuccess() {
		rt.Abortf(e, msg, args...)
}`

core/vm/exit_code/exit_code.hpp
wer1st: Why? When spec actor aborts anything, it is aborted anyway.
If we allow to keep the current error type for non-VMExitode, then this error could be changed later.
But when we cast it to VMAbortExitCode - it will be not changed

turuslan: `outcome::result` uses `std::system_error`, which combines enum value and enum id (category), so it can distinguish same enum values from different enums.
Vm should only handle `VM(Abort)ExitCode` and `VMFatal` (can't `static_cast` to `VM(Abort)ExitCode`) and avoid `static_cast(error.value())` for all other errors (e.g. `AmtError`, `HamtError`, ...), otherwise they will map to some meaningless `VMAbortExitCode` value

core/vm/exit_code/exit_code.hpp
turuslan: `Abortf` accepts `ExitCode`, not `error`

wer1st: Done

core/vm/exit_code/exit_code.hpp
wer1st: Done

turuslan (approved):

Alexey-N-Chernyshov (approved):

PR #341, opened 2021-02-25
Feature/node improvements

Alexey-N-Chernyshov:
### Description of the Change
API version
Add drand period check and default value
Add logs

### DIFF
--- a/core/node/main/config.hpp
+++ b/core/node/main/config.hpp
@@ -37,7 +37,7 @@ namespace fc::node {
     /** Drand genesis time in seconds */
     int64_t drand_genesis = 0;
     /** Drand round time in seconds */
-    int64_t drand_period = 1;
+    int64_t drand_period = 0;
     unsigned beaconizer_cache_size = 100;
 
     Config();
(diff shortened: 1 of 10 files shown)


core/node/main/config.hpp
```diff
@@ -34,8 +34,10 @@ namespace fc::node {
     // drand config
     std::vector<std::string> drand_servers;
     BlsPublicKey drand_bls_pubkey;
+    /** Drand genesis time in seconds */
     int64_t drand_genesis = 0;
-    int64_t drand_period = 0;
+    /** Drand round time in seconds */
+    int64_t drand_period = 1;
```
Alexey-N-Chernyshov, line 40: 0 is invalid period. Lotus default is 1 (see https://github.com/filecoin-project/lotus/blob/master/chain/gen/gen.go#L250)

Alexey-N-Chernyshov: Decided to keep 0 as not set.

Alexey-N-Chernyshov:
merged

PR #346, opened 2021-03-09
Feature/api update

Alexey-N-Chernyshov:
### Description of the Change
API JSON serialisation conforms Lotus.


core/api/rpc/wsc.cpp
```diff
@@ -6,14 +6,17 @@
 #include "api/rpc/wsc.hpp"
 
 #include "api/rpc/json.hpp"
-#include "api/visit.hpp"
```
ortyomka, line 9: Why is it removed?

Alexey-N-Chernyshov: It is present and used in `wsc.hpp`.

PR #356, opened 2021-03-18
cgo 3

turuslan:
- hamt 3 encoding/decoding
- init actor 3 map
- cgo 3

### DIFF
--- a/core/adt/map.hpp
+++ b/core/adt/map.hpp
@@ -26,16 +26,17 @@ namespace fc::adt {
   /// Strongly typed hamt wrapper
   template <typename Value,
             typename Keyer = StringKeyer,
-            size_t bit_width = storage::hamt::kDefaultBitWidth>
+            size_t bit_width = storage::hamt::kDefaultBitWidth,
+            bool v3 = false>
   struct Map {
     using Key = typename Keyer::Key;
     using Visitor =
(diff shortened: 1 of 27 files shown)


wer1st (approved):

core/adt/map.hpp
```diff
@@ -87,14 +88,20 @@ namespace fc::adt {
     mutable Hamt hamt;
   };
 
+  template <typename Value,
```
Alexey-N-Chernyshov, line 91: Please, add a comment that it is a map over Hamt v3.

core/storage/hamt/hamt.hpp
```diff
@@ -62,67 +67,11 @@ namespace fc::storage::hamt {
     using Item = boost::variant<CID, Ptr, Leaf>;
 
     std::map<size_t, Item> items;
+    boost::optional<bool> v3;
```
Alexey-N-Chernyshov, line 70: May be add `version enum` in case we have v4 and etc. Or leave it is a problem of future.

core/adt/map.hpp
Alexey-N-Chernyshov: And for Map v1-v2.
Links for references will be very useful.

Alexey-N-Chernyshov (approved):

PR #359, opened 2021-03-22
params and policies configuration through profiles defined on startup

Alexey-N-Chernyshov:

### Description of the Change
Add profiles for policies and parameters configuration.

### Possible Drawbacks 
- breaks naming conventions kConstants are not actually constants.
- introduces grobal variables 

### Alternate Designs *[optional]*
- create structs for policies and parameters and pass them instead of global vars.


core/const.cpp
```diff
@@ -4,9 +4,97 @@
...
+#include "vm/actor/builtin/types/verified_registry/policy.hpp"
 
 namespace fc {
-  DEFINE(kConsensusMinerMinPower){BigInt{1} << 40};
+  using primitives::sector::RegisteredSealProof;
+
+  /**
+   * Class for profile name validation with boost::program_options
+   */
+  class Profile : public std::string {};
```
wer1st, line 21: Why not `using Profile = std::string`?

core/vm/actor/builtin/types/market/policy.hpp
```diff
@@ -25,7 +25,7 @@ namespace fc::vm::actor::builtin::types::market {
                                           * TokenAmount{"1000000000000000000"}};
 
   /// The number of blocks between payouts for deals
-  constexpr EpochDuration kDealUpdatesInterval{fc::kEpochsInDay};
+  static EpochDuration kDealUpdatesInterval = fc::kEpochsInDay;
```
wer1st, line 28: Why static? constexpr looks better

wer1st (approved):

core/const.cpp
Alexey-N-Chernyshov: fixed

core/vm/actor/builtin/types/market/policy.hpp
Alexey-N-Chernyshov: It is not `const` anymore, changed to extern.

core/const.cpp
Alexey-N-Chernyshov: Actually `using Profile = std::string` creates type alias, but not the new type required for validation.

turuslan (approved):

wer1st (approved):

PR #375, opened 2021-04-02
FSM - reschedule event in case FROM state is not found

Alexey-N-Chernyshov:
See test for details

### Description of the Change
Fix FSM, see test SendBeforeConditionMet.
Previously the event that doesn't have corresponding from state was dropped.


core/fsm/fsm.hpp
```diff
@@ -440,6 +442,12 @@ namespace fc::fsm {
                                source_state,          // source state
                                resulting_state.get());  // destination state
         }
+      } else {
+        // there were no rule for transition put event in queue in case it can
+        // be handled when initial state is changed
+        std::lock_guard lock(event_queue_mutex_);
+        event_queue_.push(event_pair);
+        tickAsync();
```
Alexey-N-Chernyshov, line 450: Could it start endless loop in case there is only 'not ready to serve' events? Possible solutions - not to schedule tick or schedule delayed event.

turuslan (approved):

wer1st (approved):

PR #388, opened 2021-04-22
Feature/retrieval markets

Alexey-N-Chernyshov:
### Benefits
Add retrieval client
Update API


core/api/make.cpp
```diff
@@ -345,8 +356,80 @@ namespace fc::api {
...
+          if (piece_cid.has_value()) {
+            peers.erase(std::remove_if(peers.begin(),
+                                       peers.end(),
+                                       [&piece_cid](auto &peer) {
+                                         return peer.piece != *piece_cid;
+                                       }),
+                        peers.end());
+          }
+
+          std::mutex mutex;
```
turuslan, line 373: Use `AsyncAll<>`

core/api/make.cpp
```diff
@@ -65,6 +72,8 @@ namespace fc::api {
   using vm::state::StateTreeImpl;
   using vm::version::getNetworkVersion;
 
+  static Logger logger = common::createLogger("Full node API");
```
wer1st, line 75: Could it be const?

wer1st (approved):

core/api/make.cpp
Alexey-N-Chernyshov: Yep.

core/api/make.cpp
Alexey-N-Chernyshov: I don't see benefits from `AsyncAll<>` in this case. First, it fails if one of queries fails. Then, it cannot store call parameter that is being used here. The benefit might be in performance of `std::atomic` over `mutex`.  I don't think it is  relevant here so I prefer to leave it as is. The other option might be is to use the same approach as in 'AsyncAll' if performance is an issue. Third option is to rework `AsyncAll`, so it can be utilised here. Forth option is to add more wrappers for parameters and error handling and reuse 'AsyncAll'. I see the last option as undesirable since it reduces readability.

core/api/make.cpp
```diff
@@ -345,8 +356,72 @@ namespace fc::api {
...
+                  }
+                }
+              });
+          for (const auto &peer : peers) {
+            OUTCOME_CB1(retrieval_market_client->query(
+                peer, {root_cid, {piece_cid}}, waiter->on(peer)));
+          }
+          waiter->finalize();
+
+          cb(result);
```
turuslan, line 422: `cb` must be called from callback

core/api/make.cpp
```diff
@@ -345,8 +356,72 @@ namespace fc::api {
...
+                                       [&piece_cid](auto &peer) {
+                                         return peer.piece != *piece_cid;
+                                       }),
+                        peers.end());
+          }
+
+          std::vector<QueryOffer> result;
+          auto waiter = std::make_shared<
+              AsyncWaiter<RetrievalPeer, outcome::result<QueryResponse>>>(
+              [=, &result](auto all_calls) {
```
turuslan, line 376: `result` must not be captured but declared in callback

turuslan (approved):

PR #403, opened 2021-06-07
Compacter doesn't delete parent states for tipsets with height > main…

Alexey-N-Chernyshov:
… branch height

### Description of the Change
Check height > main branch height and save them in compacter. This allows restart after long pause and compaction.

### DIFF
--- a/core/storage/compacter/compacter.cpp
+++ b/core/storage/compacter/compacter.cpp
@@ -234,7 +234,7 @@ namespace fc::storage::compacter {
       }
       ts = ts_load->load(ts->getParents()).value();
     }
-    auto head_res{interpreter_cache->get(head->key).value()};
+    const auto head_res{interpreter_cache->get(head->key).value()};
     queue->push(*asBlake(head_res.state_root));
     queue->push(*asBlake(head_res.message_receipts));
     for (auto &branch : *ts_branches) {
--- a/core/storage/compacter/compacter.cpp
+++ b/core/storage/compacter/compacter.cpp
@@ -234,6 +234,9 @@ namespace fc::storage::compacter {
       }
       ts = ts_load->load(ts->getParents()).value();
     }
+    auto head_res{interpreter_cache->get(head->key).value()};
+    queue->push(*asBlake(head_res.state_root));
+    queue->push(*asBlake(head_res.message_receipts));
     for (auto &branch : *ts_branches) {
       if (branch == ts_main) {
         continue;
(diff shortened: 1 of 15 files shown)


core/storage/compacter/compacter.cpp
```diff
@@ -244,11 +247,14 @@ namespace fc::storage::compacter {
           primitives::tipset::put(nullptr, put_block_header, block);
           queue->push(*asBlake(block.messages));
         }
+        // if cache has tipsets with height > main_height, save them
+        auto tipset_res{interpreter_cache->tryGet(ts->key)};
+        if (tipset_res.has_value() && (ts->height() > head->height())) {
+          queue->push(*asBlake(tipset_res->value().state_root));
+          queue->push(*asBlake(tipset_res->value().message_receipts));
+        }
```
turuslan, line 255:
```suggestion
        if (ts->height() >= head->height()) {
          if (auto res{interpreter_cache->tryGet(ts->key)}; res && *res) {
            queue->push(*asBlake(res->value().state_root));
            queue->push(*asBlake(res->value().message_receipts));
          }
        }
```

turuslan (approved):

core/storage/compacter/compacter.cpp
```diff
@@ -234,6 +234,9 @@ namespace fc::storage::compacter {
       }
       ts = ts_load->load(ts->getParents()).value();
     }
+    auto head_res{interpreter_cache->get(head->key).value()};
```
wer1st, line 237:
```suggestion
    const auto head_res{interpreter_cache->get(head->key).value()};
```

wer1st, line 237: Could it has an error?

test/core/vm/interpreter/interpreter_cache_test.cpp
```diff
@@ -0,0 +1,38 @@
...
+    Result result{.state_root = state_root,
+                  .message_receipts = "010001020003"_cid,
+                  .weight = 1};
+    TipsetKey tipset_key;
+    interpreter_cache->set(tipset_key, result);
+
+    EXPECT_CALL(*ipld, get(Eq(asBlake(state_root)), Eq(nullptr)))
+        .WillOnce(Return(false));
+
+    auto res = interpreter_cache->tryGet(tipset_key);
```
wer1st, line 35:
```suggestion
    const auto res = interpreter_cache->tryGet(tipset_key);
```

wer1st (approved):

core/storage/compacter/compacter.cpp
Alexey-N-Chernyshov: No, chain head must be interpreted and cached.

wer1st: ok

turuslan: No, program keeps `head` consistent/correct, so there should not be error

PR #417, opened 2021-07-03
aggregate seal

turuslan:
- update "filecoin-ffi".
- generate and verify aggregate seal proof, with test.
- "proof_engine_impl.hpp" was not formatted.


core/proofs/impl/proof_engine_impl.cpp
```diff
@@ -8,12 +8,21 @@
...
 #include <thread>
 
+#include "codec/uvarint.hpp"
 #include "common/ffi.hpp"
-#include "primitives/address/address_codec.hpp"
 #include "primitives/cid/comm_cid.hpp"
 #include "proofs/proofs_error.hpp"
 #include "sector_storage/zerocomm/zerocomm.hpp"
 
+#define PROOFS_TRY(label)                                     \
```
ortyomka, line 17: Maybe add variable as second parameter, otherwise MACRO requires 'res_ptr' variable

ortyomka (approved):

core/proofs/impl/proof_engine_impl.cpp
Alexey-N-Chernyshov: Or rewrite it as a function.

Alexey-N-Chernyshov (approved):

PR #428, opened 2021-07-28
Feature/miner actor partition

wer1st:
### Description of the Change
Partition v0, v2, v3


core/primitives/rle_bitset/rle_bitset.hpp
```diff
@@ -73,7 +73,23 @@ namespace fc::primitives {
...
+          result.insert(i);
+        }
+      }
+      return result;
+    }
+
+    inline RleBitset slice(uint64_t start, uint64_t count) const {
+      std::vector<uint64_t> temp{begin(), end()};
+      temp = {temp.begin() + start, temp.begin() + count};
+      return RleBitset(temp.begin(), temp.end());
```
turuslan, line 89:
```suggestion
      assert(start + count <= size());
      auto it{std::next(begin(), start)};
      return {it, std::next(it, count)};
```

test/core/vm/actor/builtin/types/miner/test_utils.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+namespace fc::vm::actor::builtin {
+  using primitives::ChainEpoch;
+  using primitives::DealWeight;
+  using primitives::SectorNumber;
+  using primitives::TokenAmount;
+  using types::miner::ExpirationQueue;
+  using types::miner::SectorOnChainInfo;
+
+  class Utils {
```
turuslan, line 22: this class should be removed, prefer namespace

wer1st: Namespace is not suitable in this case because we have duplicate symbol error during compilation.
The way to use namespace is to move function implementation to .cpp file. Or just leave as it is.
What do you prefer?

turuslan (approved):

test/core/vm/actor/builtin/types/miner/test_utils.hpp
wer1st: As I understand, we leave it as is

turuslan: duplicate symbols are resolved with `inline` or `constexpr`

wer1st: thank you, fixed

test/core/vm/actor/builtin/types/miner/partition_test_v0.cpp
```diff
@@ -0,0 +1,743 @@
...
+
+    std::vector<SectorOnChainInfo> sectors;
+    SectorSize ssize{static_cast<uint64_t>(32) << 30};
+    QuantSpec quant{4, 1};
+
+    Partition partition;
+    std::vector<ExpectExpirationGroup> groups;
+  };
+
+  TEST_F(PartitionTestV0, AddsSectorsAndReportsSectorStats) {
```
Alexey-N-Chernyshov, line 246: Please, add test descriptions here and further.

core/primitives/rle_bitset/rle_bitset.hpp
```diff
@@ -83,6 +99,17 @@ namespace fc::primitives {
       }
       return result;
     }
+
+    inline bool containsAny(const RleBitset &other) const {
```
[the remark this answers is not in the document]
wer1st: We stay with `containsAny` because it is consistent with SpecActors

Alexey-N-Chernyshov (approved):

PR #437, opened 2021-08-09
deposit pledge

turuslan:
- `StateMinerInitialPledgeCollateral`
- `StateMinerPreCommitDepositForPower`
- fix `MessageVisitor` bug

### DIFF
--- a/core/vm/actor/builtin/v5/market/validate.hpp
+++ b/core/vm/actor/builtin/v5/market/validate.hpp
@@ -1,0 +1,54 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include "vm/actor/builtin/states/market/market_actor_state.hpp"
+#include "vm/actor/builtin/types/market/policy.hpp"
+#include "vm/exit_code/exit_code.hpp"
+
+namespace fc::vm::actor::builtin::v5::market {
+  using primitives::DealId;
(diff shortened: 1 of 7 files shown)


core/vm/actor/builtin/v5/market/validate.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include "vm/actor/builtin/types/market/policy.hpp"
+#include "vm/exit_code/exit_code.hpp"
+
+namespace fc::vm::actor::builtin::v5::market {
+  using primitives::DealId;
+  using primitives::SpaceTime;
+  using primitives::address::Address;
+  using primitives::piece::PaddedPieceSize;
+
+  struct validate_RESULT {
```
wer1st, line 18:
```suggestion
  struct ValidateResult {
```

wer1st (approved):

core/vm/actor/builtin/v5/market/validate.hpp
turuslan: i don't like `ValidateResult` because it changes `validate` and doesn't split it from "result".
this is like named tuple, used only by this function.
may rename if you want and will suggest name (please confirm to rename to `ValidateResult`).

Alexey-N-Chernyshov (approved):

core/vm/actor/builtin/v5/market/validate.hpp
wer1st: I think that we should to rename it, because it is a struct after all.

PR #439, opened 2021-08-11
fix blocksync

Alexey-N-Chernyshov:
### Description of the Change
From libp2p

     * @note caller should maintain validity of an input buffer until callback
     * is executed. It is usually done with either wrapping buffer as shared
     * pointer, or having buffer as part of some class/struct, and using
     * enable_shared_from_this()

### DIFF
--- a/core/common/libp2p/cbor_stream.hpp
+++ b/core/common/libp2p/cbor_stream.hpp
@@ -41,18 +41,23 @@ namespace fc::common::libp2p {
       });
     }
 
-    /// Write bytes of cbor object
-    void writeRaw(gsl::span<const uint8_t> input, WriteCallbackFunc cb);
+    /**
+     * Write bytes of cbor object
+     *
+     * @param input - shared pointer to buffer that must be alive until libp2p
+     * callback in stream::write() is called
+     */
+    void writeRaw(std::shared_ptr<Buffer> input, WriteCallbackFunc cb);
 
     /// Write cbor object
     template <typename T>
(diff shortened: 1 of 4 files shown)


turuslan (approved):

wer1st (approved):

core/node/blocksync_request.cpp
```diff
@@ -275,14 +275,15 @@ namespace fc::sync::blocksync {
           depth = kMaxDepth;
         }
 
-        auto binary_request = codec::cbor::encode<Request>(
+        const auto maybe_binary_request = codec::cbor::encode<Request>(
```
turuslan, line 278:
```suggestion
        auto maybe_binary_request = codec::cbor::encode<Request>(
```
don't `const` what needs to be `move`

core/common/libp2p/cbor_stream.hpp
```diff
@@ -41,18 +41,22 @@ namespace fc::common::libp2p {
...
+     */
+    void writeRaw(std::shared_ptr<Buffer> input, WriteCallbackFunc cb);
 
     /// Write cbor object
     template <typename T>
     void write(const T &value, WriteCallbackFunc cb) {
-      auto encoded = codec::cbor::encode(value);
-      if (!encoded) {
-        return cb(encoded.error());
+      const auto maybe_encoded = codec::cbor::encode(value);
```
turuslan, line 55: don't `const` what needs `move`

PR #459, opened 2021-09-13
Clang-tidy fixes.

Alexey-N-Chernyshov:

### Description of the Change
More clang-tidy fixes.
Explicit constructors, const ref parameters, getter for private class methods and etc.


core/adt/channel.hpp
```diff
@@ -108,7 +108,10 @@ namespace fc::adt {
         return false;
       }
       if (which<Handler>(state)) {
-        boost::get<Handler>(state)({});
+        try {
```
Alexey-N-Chernyshov, line 111: Here was an unhandled exception in destructor.

turuslan: silently suppressing exceptions is not good.

also, there were no exception, because of type check in `if` in previous line.

may rewrite as `if (auto handler{boost::get<Handler>(&state)}) { handler({});` }

core/vm/interpreter/interpreter.hpp
```diff
@@ -32,7 +32,7 @@ namespace fc::vm::interpreter {
 
   struct InterpreterCache {
     struct Key {
-      Key(const TipsetKey &tsk);
+      explicit Key(const TipsetKey &tsk);
```
turuslan, line 35: explicit constructor looks ugly.  
this struct was intended to avoid duplicating methods (has/get/set/remove) which can accept either `TipsetKey` or reuse allocated `Key` in subsequent calls (eg cached interpreter, not sure if there is any meaning in reuse).
may remove this struct and change `Key` usages to `TipsetKey`.

core/api/full_node/make.cpp
```diff
@@ -1074,8 +1088,8 @@ namespace fc::api {
     };
 
     // TODO(artyom-yurin): FIL-165 implement method
-    api->StateSectorPreCommitInfo = std::function<decltype(
-        api->StateSectorPreCommitInfo)::FunctionSignature>{};
+    api->StateSectorPreCommitInfo = std::function<
+        decltype(api->StateSectorPreCommitInfo)::FunctionSignature>{};
```
turuslan, line 1092:
```suggestion
```

core/vm/runtime/runtime.hpp
```diff
@@ -253,43 +253,43 @@ namespace fc::vm::runtime {
...
     }
 
     /// Send with typed result R
     template <typename R>
-    outcome::result<R> sendR(Address to_address,
+    outcome::result<R> sendR(const Address &to_address,
                              MethodNumber method_number,
                              const MethodParams &params,
-                             BigInt value) {
+                             const BigInt &value) {
```
wer1st, line 272: The same changes for send() method on line 126.
And maybe we want to use TokenAmount instead of BigInt here

wer1st (approved):

core/adt/channel.hpp
turuslan: silently ignoring exceptions is bad.
if logging `catch (...)` is difficult, maybe should add `io_context` to move potentially throwing call from destructor.
or remove `catch` and see if any handler will throw here (they should not, because they work with outcomes).

Alexey-N-Chernyshov: Updated

turuslan (approved):

PR #462, opened 2021-09-19
update onDealSectorCommitted

turuslan:
- outcome callback.
- detect missed precommit and commit messages.
- use `PreCommitBatch` and `ProveCommitAggregate` messages.
- remove unnecessary optional.
- simplify `makeEmptyMinerState `, `makeEmptyDeadlines`, `makeEmptyDeadline`.
- update test.

### DIFF
--- a/core/vm/actor/builtin/states/miner/miner_actor_state.cpp
+++ b/core/vm/actor/builtin/states/miner/miner_actor_state.cpp
@@ -454,15 +454,11 @@ namespace fc::vm::actor::builtin::states {
         .total_faulty_power = total_faulty_power};
   }
 
-  outcome::result<MinerActorStatePtr> makeEmptyMinerState(
-      const Runtime &runtime) {
-    const auto version = runtime.getActorVersion();
-    const auto ipld = runtime.getIpfsDatastore();
-
-    MinerActorStatePtr state{version};
+  outcome::result<MinerActorStatePtr> makeEmptyMinerState(const IpldPtr &ipld) {
+    MinerActorStatePtr state{ipld->actor_version};
     cbor_blake::cbLoadT(ipld, state);
 
-    if (version < ActorVersion::kVersion3) {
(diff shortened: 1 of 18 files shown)


core/vm/actor/builtin/states/miner/miner_actor_state.cpp
```diff
@@ -454,15 +454,11 @@ namespace fc::vm::actor::builtin::states {
         .total_faulty_power = total_faulty_power};
   }
 
-  outcome::result<MinerActorStatePtr> makeEmptyMinerState(
-      const Runtime &runtime) {
-    const auto version = runtime.getActorVersion();
-    const auto ipld = runtime.getIpfsDatastore();
-
-    MinerActorStatePtr state{version};
+  outcome::result<MinerActorStatePtr> makeEmptyMinerState(const IpldPtr &ipld) {
```
wer1st, line 457: I don't see any places in your code where you need this function. It is always used by context that has "runtime" to provide it.
Why have you decided to change it?
Moreover, we could have an error if someone provides ipld without correct actor version.

core/common/std_function_signature.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+namespace fc {
+  template <typename T>
+  struct std_function_signature {};
+  template <typename T>
+  struct std_function_signature<std::function<T>> {
+    using type = T;
+  };
+  template <typename T>
+  using std_function_signature_t = typename std_function_signature<T>::type;
```
wer1st, line 18:
```suggestion
  template <typename T>
  struct std_function_signature {};
  
  template <typename T>
  struct std_function_signature<std::function<T>> {
    using type = T;
  };
  
  template <typename T>
  using std_function_signature_t = typename std_function_signature<T>::type;
```

wer1st (approved):

Alexey-N-Chernyshov (approved):

PR #473, opened 2021-10-01
mpool bls cache size

turuslan:


test/core/storage/mpool/message_pool_test.cpp
```diff
@@ -79,6 +79,7 @@ namespace fc::storage::mpool {
          MessagePool::create(
              {ipld, nullptr, nullptr, ts_load, interpreter_cache},
              nullptr,
+             1000,
```
wer1st, line 82: mpool_bls_cache_size ?

core/node/main/config.hpp
```diff
@@ -47,6 +47,8 @@ namespace fc::node {
     /** Node default key path */
     boost::optional<std::string> wallet_default_key_path;
 
+    size_t mpool_bls_cache_size{1000};
```
wer1st, line 50:
```suggestion
    constexpr size_t mpool_bls_cache_size{1000};
```

test/core/storage/mpool/message_pool_test.cpp
turuslan: test works with any value

core/node/main/config.hpp
turuslan: no, that's mutable config field with default value

wer1st: Ok

wer1st (approved):

Alexey-N-Chernyshov (approved):

PR #475, opened 2021-10-06
Miner fix

ortyomka:
Signed-off-by: ortyomka <iurin.art@gmail.com>

### Description of the Change
- Fix ChainGetTipSetByHeight
- Separate Chan and Client Return contexts. Otherwise channels can block return API and it would be stacked.
- Add migration from old metadata
- Fix WindowPostScheduler
- Added v5 actors types


core/api/rpc/wsc.hpp
```diff
@@ -53,9 +53,12 @@ namespace fc::api::rpc {
 
    private:
     std::thread thread;
+    std::thread thread_chan;
```
turuslan, line 56: replace with `IoThread`

PR #485, opened 2021-10-20
Buffer remove

ortyomka:
### Description of the Change
- Remove Buffer from the system

### DIFF
--- a/core/adt/set.hpp
+++ b/core/adt/set.hpp
@@ -18,7 +18,13 @@ namespace fc::adt {
   }
 
   template <typename Keyer, size_t bit_width = storage::hamt::kDefaultBitWidth>
-  struct Set : Map<SetValue, Keyer, bit_width> {};
+  struct Set : Map<SetValue, Keyer, bit_width> {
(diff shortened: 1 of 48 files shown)


wer1st:
> > Separate methods are better, they can be forward/declared elsewhere.  
> 
> Inheriting is inconvenient, `struct Bytes { public: vector v; }` may be more convenient.  
> 
> 

Disagree, if method is a part of Bytes, it goes everywhere when we use Bytes.
Vector as a class member is the same approach as removed Buffer and it is not convenient. In this case you need to call bytes.vector.method() or write a lot of wrapper methods. Moreover, if we select to call vector's methods from bytes.vector and new mrthods like append() from bytes: bytes.method(), it is inconsistent of usage this class.

From the other side, if we use inheritance we can call vector's methods directly as bytes.method() and call new methods the same way that is consistent. Also there are no problems with convertation between Bytes and vector that allows us to use this class more smoothly in many places.
Look at RleBitset that inherits std::set, we got very useful and convenient class from inheritance approach.

PR #496, opened 2021-11-02
paych maker

turuslan:
- extract `PaychGet` from `PaymentChannelManager` into `PaychMaker`.
- prevent duplicate channels with same from-to.
- batch pending `PaychGet` to single message.
- sending one message at a time and waiting before sending next.
- save relevant info to db.
- fix `StateWaitMsg` lock.

- may rename, e.g. ~router.

### DIFF
--- a/core/api/impl/paych_get.hpp
+++ b/core/api/impl/paych_get.hpp
@@ -1,0 +1,22 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include "api/full_node/node_api.hpp"
+#include "payment_channel_manager/impl/maker.hpp"
+
+namespace fc::api {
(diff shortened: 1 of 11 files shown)


core/api/impl/paych_get.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+namespace fc::api {
+  inline void implPaychGet(
+      const std::shared_ptr<FullNodeApi> &api,
+      const std::shared_ptr<paych_maker::PaychMaker> &maker) {
+    api->PaychGet = [=](auto &&cb,
+                        const Address &from,
+                        const Address &to,
+                        const TokenAmount &amount) {
+      maker->make({from, to}, amount, std::move(cb));
```
Alexey-N-Chernyshov, line 19: `const ref` cb moved, has no effect here. You can add `mutable` to lambda to fix it.

Alexey-N-Chernyshov (approved):

PR #497, opened 2021-11-03
fix randomness

turuslan:
- beacon randomness for network v14, uses drand schedule.
- fix interopnet max version.
- fix randomness api lock.

### DIFF
--- a/core/api/full_node/make.cpp
+++ b/core/api/full_node/make.cpp
@@ -335,6 +335,7 @@ namespace fc::api {
             auto &entropy) -> outcome::result<Randomness> {
       std::unique_lock ts_lock{*env_context.ts_branches_mutex};
       OUTCOME_TRY(ts_branch, TsBranch::make(ts_load, tipset_key, ts_main));
+      ts_lock.unlock();
       return env_context.randomness->getRandomnessFromBeacon(
           ts_branch, tag, epoch, entropy);
     };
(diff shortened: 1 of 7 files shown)


core/vm/runtime/impl/tipset_randomness.cpp
```diff
@@ -5,42 +5,77 @@
...
+                     network < NetworkVersion::kVersion13));
     OUTCOME_TRY(ts, ts_load->lazyLoad(it.second->second));
     ts_lock.unlock();
 
     return crypto::randomness::drawRandomness(
         ts->getMinTicketBlock().ticket->bytes, tag, epoch, seed);
   }
 
+  inline outcome::result<BeaconEntry> extractBeaconEntryForEpoch(
+      TsLoadPtr ts_load,
```
Alexey-N-Chernyshov, line 41: why not const ptr?

Alexey-N-Chernyshov (approved):

core/api/full_node/make.cpp
```diff
@@ -335,6 +335,7 @@ namespace fc::api {
             auto &entropy) -> outcome::result<Randomness> {
       std::unique_lock ts_lock{*env_context.ts_branches_mutex};
       OUTCOME_TRY(ts_branch, TsBranch::make(ts_load, tipset_key, ts_main));
+      ts_lock.unlock();
```
wer1st, line 338: Why if it should be free after next line?

wer1st (approved):

core/api/full_node/make.cpp
turuslan: because function called on next line (before destructor) needs `shared_lock` on this mutex.

PR #541, opened 2021-11-30
Miner deal fixes

ortyomka:
### Description of the Change

- New Stored Ask (RAII)
- Local JWT library
- New method `MarketListIncompleteDeals`
- Fix Fd behavior in Read Piece
- Remove duplicate types


core/miner/main/main.cpp
```diff
@@ -96,6 +100,24 @@ namespace fc {
     auto join(const std::string &path) const {
       return (repo_path / path).string();
     }
+
+    const std::string &localIp() const {
```
turuslan, line 104: move inline function to core/common header (reuse in node main)

turuslan (approved):

core/miner/main/main.cpp
ortyomka: Done

PR #552, opened 2021-12-10
fetch msg

turuslan:
- `PeerHeight` remembers peers' heights.
- `FetchMsg` queues and fetches messages.
- sync_job.cpp fetches messages and blocks, doesn't wait for messages until vm execution.
- split sync_job.cpp long blocking loop.
- fix `BlocksyncRequest`.
- `PeerDisconnected` event.

note: may prefetch more messages for vm (not implemented).


core/node/blocksync_request.hpp
```diff
@@ -48,6 +48,8 @@ namespace fc::sync::blocksync {
 
       /// All their meta/messages is also available
       bool messages_stored = false;
+
+      bool messages_only{};
```
wer1st, line 52: Please define it more clearly. Use directly `true` or `false`

wer1st (approved):

Alexey-N-Chernyshov (approved):

PR #561, opened 2021-12-29
buffered car append

turuslan:


core/storage/ipld/cids_ipld.cpp
```diff
@@ -179,22 +183,59 @@ namespace fc::storage::ipld {
...
+    std::shared_lock lock{car_flush_mutex};
+    const auto it{car_queue.find(row.offset.value())};
+    if (it == car_queue.end()) {
+      return false;
+    }
+    BytesIn item{it->second};
+    BytesIn input;
+    const CbCid *hash;
+    if (codec::uvarint::readBytes(input, item)
+        && codec::cbor::light_reader::readCborBlake(hash, input)) {
```
wer1st, line 213: I suggest to rework `readCborBlake()` from out parameter to return based approach.
`const CbCid *hash = readCborBlake(input)`
`if (hash != nullptr) { ... }`

wer1st: @Alexey-N-Chernyshov you have TODO for that in `readCborBlake()`
What do you think?

Alexey-N-Chernyshov: Yes, I prefer not to use out parameters, and here the function like `bool isCborBlakeObject(input)` would be more meaningful to check if input has valid object.
But I think the point of discussion is out of PR scope.

Alexey-N-Chernyshov (approved):

core/storage/ipld/cids_ipld.cpp
wer1st: I agree that we could fix it in separate ticket

wer1st (approved):

wer1st (approved):

Alexey-N-Chernyshov (approved):

PR #565, opened 2022-01-12
Feature/wallet

wer1st:
### Description of the Change
Move wallet API methods to new Wallet entity.
Introduce MultiWallet conception that allows us to work with multiple wallets by using one multi wallet.
LedgerWallet has to be added in MultiWallet.


core/api/wallet/local_wallet.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+namespace fc::api {
+  using vm::state::StateTreeImpl;
+
+  void LocalWallet::fillLocalWalletApi(
+      const std::shared_ptr<WalletApi> &api,
+      const std::shared_ptr<KeyStore> &key_store,
+      const IpldPtr &ipld,
+      const std::shared_ptr<OneKey> &wallet_default_address) {
+    const StateTreeImpl state_tree{ipld};
```
turuslan, line 22:
```suggestion
      const std::function<outcome::result<TipsetContext>()> &get_ts_ctx,
      const std::shared_ptr<OneKey> &wallet_default_address) {
```
can move `tipsetContext` lambda from `api/full_node/make.cpp` to `builder.cpp` and reuse for both/all `makeImpl/fill...`

turuslan (approved):

Alexey-N-Chernyshov (approved):

PR #567, opened 2022-01-21
Ledger wallet

wer1st:
### Description of the Change
Implementation of Ledger Wallet

### DIFF
--- a/core/api/wallet/ledger.cpp
+++ b/core/api/wallet/ledger.cpp
@@ -1,0 +1,118 @@
+/**
+ * Copyright Soramitsu Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "api/wallet/ledger.hpp"
+
+#include <algorithm>
+#include "api/rpc/json.hpp"
+#include "codec/json/json.hpp"
+#include "common/bytes_cow.hpp"
+#include "common/error_text.hpp"
+#include "cpp-ledger/filecoin/ledger_filecoin.hpp"
+#include "primitives/address/address_codec.hpp"
+
+namespace fc::api {
+  using ledger::kPathLength;
+  using ledger::filecoin::LedgerFilecoinManager;
+  using primitives::address::decodeFromString;
+
+  const auto app_error = ERROR_TEXT("cannot open Ledger Filecoin App");
+
+  Ledger::Ledger(const MapPtr &store) : store(store) {}
+
+  bool Ledger::Has(const Address &address) const {
+    return store->contains(encode(address));
+  }
+
+  outcome::result<Signature> Ledger::Sign(const Address &address,
+                                          const Bytes &data) const {
+    OUTCOME_TRY(value, store->get(encode(address)));
+    OUTCOME_TRY(j_file, codec::json::parse(value));
+    OUTCOME_TRY(ledger_key_info, decode<LedgerKeyInfo>(j_file));
+
+    const auto [app, err1] = LedgerFilecoinManager::FindLedgerFilecoinApp();
+    if (err1 != std::nullopt) {
+      return app_error;
+    }
+
+    const auto [sign, err2] = app->SignSECP256K1(ledger_key_info.path, data);
+    app->Close();
+    if (err2 != std::nullopt) {
+      return ERROR_TEXT("sign error");
+    }
+
+    OUTCOME_TRY(signature,
+                Blob<crypto::secp256k1::kSignatureLength>::fromSpan(
+                    sign.SignatureBytes()));
+
+    return signature;
+  }
+
+  outcome::result<Address> Ledger::ImportKey(
+      const LedgerKeyInfo &key_info) const {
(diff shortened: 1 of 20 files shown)


core/api/wallet/ledger.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+    outcome::result<Signature> Sign(const Address &address,
+                                    const Bytes &data) const;
+
+    outcome::result<Address> ImportKey(const LedgerKeyInfo &key_info) const;
+
+    outcome::result<Address> New() const;
+
+   private:
+    const MapPtr &store;
```
turuslan, line 35:
```suggestion
    MapPtr store;
```

core/api/wallet/ledger.cpp
```diff
@@ -0,0 +1,118 @@
...
+  using ledger::kPathLength;
+  using ledger::filecoin::LedgerFilecoinManager;
+  using primitives::address::decodeFromString;
+
+  const auto app_error = ERROR_TEXT("cannot open Ledger Filecoin App");
+
+  Ledger::Ledger(const MapPtr &store) : store(store) {}
+
+  bool Ledger::Has(const Address &address) const {
+    return store->contains(encode(address));
```
turuslan, line 26: Device master seed may change, so db cached address/path pair becomes invalid. `Has` may ask and compare address from device.  
Add `// note:` (or todo) about this.

core/api/wallet/ledger.cpp
```diff
@@ -0,0 +1,118 @@
...
+                    sign.SignatureBytes()));
+
+    return signature;
+  }
+
+  outcome::result<Address> Ledger::ImportKey(
+      const LedgerKeyInfo &key_info) const {
+    if (key_info.address == Address{}) {
+      return ERROR_TEXT("no address given in imported key info");
+    }
```
turuslan, line 57: Lotus supports empty address in json, but our decoder won't accept no value.
```suggestion
```

core/api/wallet/ledger.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+    if (key_info.path.size() != kPathLength) {
+      return ERROR_TEXT("bad hd path len");
+    }
+
```
turuslan, line 62: ask device to derive key-address and compare with expected from json

core/api/wallet/ledger.cpp
```diff
@@ -0,0 +1,118 @@
...
+    // Filecoin base path + new id
+    const std::vector<uint32_t> path{hdHard | 44, hdHard | 461, hdHard, 0, id};
+
+    std::string addr;
+
+    std::tie(std::ignore, std::ignore, addr, err) =
+        app->GetAddressPubKeySECP256K1(path);
+    if (err != std::nullopt) {
+      app->Close();
```
turuslan, line 102: does app support destructor?

Alexey-N-Chernyshov (approved):

core/api/wallet/ledger.hpp
wer1st: Why?

turuslan: Why not create `Ledger` object once like others (eg Keystore).  Reference to shared_ptr will dangle then.

core/api/wallet/ledger.cpp
wer1st: Yes, it could be moved to destructor

core/api/wallet/ledger.hpp
wer1st: Ledger object is created and used only inside of LedgerWallet API methods. And the current approach allows us to save some memory, because ref is always alive when Ledger object alive.
From the other hand, it is not safety if someone will use Ledger outside of LedgerWallet, for these cases I can change the code

core/api/wallet/ledger.cpp
```diff
@@ -22,8 +22,34 @@ namespace fc::api {
...
+      auto [app, err] = LedgerFilecoinManager::FindLedgerFilecoinApp();
+      if (err != std::nullopt) {
+        return app_error;
+      }
+
+      std::string addr;
+      std::tie(std::ignore, std::ignore, addr, err) =
+          app->GetAddressPubKeySECP256K1(ledger_key_info.path);
+      if (err != std::nullopt) {
+        OUTCOME_TRY(store->remove(encode(address)));
```
turuslan, line 40: may be dangerous (eg user who has two ledgers and inserted wrong one will lose keys from other device).  

don't remove.  
there is explicit api `WalletDelete` for removal
```suggestion
```

core/api/wallet/ledger.cpp
```diff
@@ -22,8 +22,34 @@ namespace fc::api {
...
+        return false;
+      }
+
+      OUTCOME_TRY(address_ledger, decodeFromString(addr));
+      if (address_ledger != address) {
+        OUTCOME_TRY(store->remove(encode(address)));
```
turuslan, line 46:
```suggestion
```

turuslan (approved):

PR #575, opened 2022-01-31
Api fix remote worker

Elestrias:


Elestrias:
Old commits here for stable version, after master
 fixes it will be removed

PR #600, opened 2022-02-17
Reconnect socket && worker ping

Elestrias:
Signed-off-by: elestrias <rus8-2002@mail.ru>

### DIFF
--- a/core/api/rpc/wsc.cpp
+++ b/core/api/rpc/wsc.cpp
@@ -37,14 +37,14 @@ namespace fc::api::rpc {
     return connect(ip, port, target, token);
   }
 
-    outcome::result<void> Client::connect(const std::string &host,
-                                          const std::string &port,
-                                          const std::string &target,
-                                          const std::string &token) {
+  outcome::result<void> Client::connect(const std::string &host,
+                                        const std::string &port,
+                                        const std::string &target,
+                                        const std::string &token) {
     boost::system::error_code ec;
     socket->next_layer().connect({boost::asio::ip::make_address(host),
-                                 boost::lexical_cast<uint16_t>(port)},
-                                ec);
+                                  boost::lexical_cast<uint16_t>(port)},
+                                 ec);
     if (ec) {
       return ec;
     }
@@ -92,19 +92,19 @@ namespace fc::api::rpc {
   }
 
   void Client::_flush() {
-    if (!writing && !write_queue.empty() && not reconnecting){
+    if (!writing && !write_queue.empty() && not reconnecting) {
       auto &[id, buffer] = write_queue.front();
       writing = true;
       socket->async_write(boost::asio::buffer(buffer.data(), buffer.size()),
-                         [=](auto &&ec, auto) {
-                           std::lock_guard lock{mutex};
-                           if (ec) {
-                             return _error(ec);
-                           }
-                           writing = false;
-                           write_queue.pop();
-                           _flush();
-                         });
+                          [=](auto &&ec, auto) {
+                            std::lock_guard lock{mutex};
+                            if (ec) {
+                              return _error(ec);
+                            }
+                            writing = false;
+                            write_queue.pop();
+                            _flush();
+                          });
     }
   }
 
@@ -189,10 +189,12 @@ namespace fc::api::rpc {
   }
 
   void Client::reconnect(int counter, std::chrono::milliseconds wait) {
-    if(reconnecting) return;
+    if (reconnecting) return;
     reconnecting = true;
-    logger_->info("Starting reconnect to {}:{}", client_data.host, client_data.port);
-    for(int i = 0; i < counter; i++){
+    bool rec_status{false};
+    logger_->info(
+        "Starting reconnect to {}:{}", client_data.host, client_data.port);
+    for (int i = 0; i < counter; i++) {
       std::this_thread::sleep_for(wait);
       socket.reset();
       socket.emplace(io);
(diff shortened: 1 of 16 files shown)


core/sector_storage/worker.hpp
```diff
@@ -142,6 +142,8 @@ namespace fc::sector_storage {
 
     virtual outcome::result<std::vector<primitives::StoragePath>>
     getAccessiblePaths() = 0;
+
+    virtual void ping(std::function<void(const bool &resp)> cb) = 0;
```
turuslan, line 146: unnecessary reference may be omited
```suggestion
    virtual void ping(std::function<void(bool)> cb) = 0;
```

Elestrias: bool is taking 1 byte of memory, why copying == referencing ?

core/api/rpc/wsc.cpp
```diff
@@ -185,4 +187,22 @@ namespace fc::api::rpc {
...
+  void Client::reconnect(int counter, std::chrono::milliseconds wait) {
+    if(reconnecting.exchange(true)) return;
+    logger_->info("Starting reconnect to {}:{}", client_data.host, client_data.port);
+    for(int i = 0; i < counter; i++){
+      std::this_thread::sleep_for(wait*(i+1));
+      auto res = connect(client_data.host,
+                         client_data.port,
+                         client_data.target,
+                         client_data.token);
+      if(!res.has_error()) {
```
ortyomka, line 200:
```suggestion
      if(not res.has_error()) {
```

core/api/rpc/wsc.cpp
```diff
@@ -185,4 +187,22 @@ namespace fc::api::rpc {
...
+
+  void Client::reconnect(int counter, std::chrono::milliseconds wait) {
+    if(reconnecting.exchange(true)) return;
+    logger_->info("Starting reconnect to {}:{}", client_data.host, client_data.port);
+    for(int i = 0; i < counter; i++){
+      std::this_thread::sleep_for(wait*(i+1));
+      auto res = connect(client_data.host,
+                         client_data.port,
+                         client_data.target,
+                         client_data.token);
```
ortyomka, line 199: Here socket would be reused. You can not do it this [way](https://github.com/boostorg/beast/issues/1778)

core/api/rpc/wsc.cpp
```diff
@@ -185,4 +187,22 @@ namespace fc::api::rpc {
...
+      if(!res.has_error()) {
+        break;
+      }
+    }
+    reconnecting.store(false);
+    logger_->info("Reconnect to {}:{} was successful", client_data.host, client_data.port);
```
ortyomka, line 205: If all N tries are failed, then logger anyway said, that reconnect successfuly

core/api/rpc/wsc.cpp
```diff
@@ -87,10 +88,11 @@ namespace fc::api::rpc {
       }
     }
     chans.clear();
+    reconnect(3, std::chrono::seconds(5));
```
ortyomka, line 91: Why do we clean everything, and only then try to fix the situation?

core/sector_storage/worker.hpp
ortyomka: If copy and reference are equal, we prefer copy

core/api/rpc/wsc.cpp
ortyomka: Also, should we recconect on each error?

core/api/rpc/wsc.hpp
```diff
@@ -57,6 +57,18 @@ namespace fc::api::rpc {
     void setup(A &api) {
       visit(api, [&](auto &m) { _setup(*this, m); });
     }
+    struct ClientData {
+      ClientData(std::string host,
+                 std::string port,
+                 std::string target,
+                 std::string token)
+          : host(host), port(port), target(target), token(token){};
+      ClientData() = default;
```
ortyomka, line 66:
```suggestion
```

core/api/rpc/wsc.cpp
turuslan: We clean, because we can't recover session (eg waiting calls and channels), so each waiting call and channel is returned and error, then api caller (eg sealing fsm, or other components) to call again or restore notifications channel.

Reconnecting on connection error allows `ChainNotify` caller to restore notification channel. Also `reconnect()` has mutex

core/sector_storage/worker.hpp
wer1st: Referencing takes 4 bytes

core/api/rpc/wsc.hpp
Elestrias: why

Elestrias:
The problem I could see is that with connection retry we can connect to that socket with  same host:port but it won't be the same process f e, instead of worker with gpu will be connected worker without gpu, problem is that we don't make changes in miner's worker info

ortyomka (approved):

core/api/rpc/wsc.cpp
```diff
@@ -87,27 +88,28 @@ namespace fc::api::rpc {
...
-                           write_queue.pop();
-                           _flush();
-                         });
+      socket->async_write(boost::asio::buffer(buffer.data(), buffer.size()),
+                          [=](auto &&ec, auto) {
+                            std::lock_guard lock{mutex};
+                            if (ec) {
+                              return _error(ec);
+                            }
+                            writing = false;
```
turuslan, line 104: move `writing = false` before `if (ec)` or add to `reconnect`, otherwise will never write anything after reconnect

turuslan (approved):

core/sector_storage/impl/manager_impl.cpp
```diff
@@ -415,7 +414,7 @@ namespace fc::sector_storage {
                                       remote,
                                       proofs);
 
-    OUTCOME_TRY(manager->addWorker(std::move(worker)));
+    //OUTCOME_TRY(manager->addWorker(std::move(worker)));
```
Elestrias, line 417:
```suggestion
    OUTCOME_TRY(manager->addWorker(std::move(worker)));
```

PR #601, opened 2022-02-18
CMake: Improve building without hunter

eigendude:
### Description of the Change

Now that https://github.com/filecoin-project/cpp-filecoin/pull/128 is merged, I've got Fuhon building without hunter again.

The approach is to delegate finding dependencies to the build system using the typical CMake approach, using find modules. I've included three find modules in this PR.

I only included find modules for three of the dependencies. The reason is because downstream, I just use the hunterized versions of dependencies. However, the hunter team made the unfortunate but easy choice of forking repos. This means that many dependencies have fallen behind the versions used by my team's app. We can't downgrade three of the packages, so I wrote CMake find modules for these three dependencies.

The final patch, for filecoin_ffi, also means that git submodules are no longer needed. This is advantageous, because git submodules cause problems for a fair number of build systems and packagers. And who hasn't forgotten `--recursive`.

It's an open question if we want CMake find modules for any additional dependencies. I could possibly add this to my List Of Things To Do (tm).

### Benefits

* Improved building of Fuhon in environments without Hunter.

### Possible Drawbacks 

CMake code is added, currently totaling about 70 lines. Normally any lines added increases the load of maintenance, but code that doesn't ship has much less of an impact.

### Alternate Designs *[optional]*

I currently use an alternative approach for the Curl dependency - I just patch out the `CONFIG` parameter, and ship the patch with the rest of my work. The exact patch is:

```
commit a136981cfaf04e7fe7f8ced38bcd9ebaf6493f91
Author: Garrett Brown <eigendebugger@gmail.com>
Date:   Tue Feb 15 10:18:31 2022 -0800

    [temp] CMake: Delegate finding CURL to the build system

diff --git a/cmake/dependencies.cmake b/cmake/dependencies.cmake
index 6c5636dd..de953113 100644
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -25,8 +25,7 @@ hunter_add_package(OpenSSL)
 find_package(OpenSSL REQUIRED)
 
 # https://hunter.readthedocs.io/en/latest/packages/pkg/CURL.html#pkg-curl
-hunter_add_package(CURL)
-find_package(CURL CONFIG REQUIRED)
+find_package(CURL REQUIRED)
 
 # https://developers.google.com/protocol-buffers/
 hunter_add_package(Protobuf)
```

By not including Curl in this PR, it increases my downstream technical debt, but not significantly.

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -46,7 +46,12 @@ find_package(Boost.DI CONFIG REQUIRED)
 
 # https://docs.hunter.sh/en/latest/packages/pkg/leveldb.html
 hunter_add_package(leveldb)
-find_package(leveldb CONFIG REQUIRED)
+if (HUNTER_ENABLED)
+  find_package(leveldb CONFIG REQUIRED)
+else()
+  find_package(leveldb REQUIRED)
+  include_directories(${LEVELDB_INCLUDE_DIRS})
+endif()
 
 # https://github.com/soramitsu/libp2p
 hunter_add_package(libp2p)
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -1,6 +1,11 @@
 # hunter dependencies
 # https://docs.hunter.sh/en/latest/packages/
 
+# Append local modules path if Hunter is not enabled
+if (NOT HUNTER_ENABLED)
+  list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
+endif()
+
 if (TESTING)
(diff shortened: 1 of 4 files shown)


turuslan (approved):

Alexey-N-Chernyshov (approved):

Alexey-N-Chernyshov:
@eigendude, thanks for your interest and cooperation.

eigendude:
@Alexey-N-Chernyshov thanks for accepting any of the patches. I get crazy ideas like bringing Filecoin to the living room and start to make them. Anything accepted means I don't have to do it twice 🙂

PR #671, opened 2022-04-27
Actor methods

wer1st:
### Description of the Change
Implementation of the common approach to call actors methods.
We don't need to call Params and Method number from the specific version anymore.
The common method interface must be always actual with the last version of actors.


core/vm/actor/builtin/v4/todo.hpp
```diff
@@ -1,25 +0,0 @@
-/**
```
turuslan, line 1: Why? It was working.  
I would keep assertion, rather than debugging unexpected/silent errors.

turuslan (approved):

Alexey-N-Chernyshov (approved):

